PR #28, opened 2022-08-12
remove legacy mapper related code

DIFSRIP:


DIFSRIP:
closed since mixed up with workflow configuration

PR #32, opened 2022-08-14
Parallelization

Templada:


DIFSRIP:
@Templada can you add your changes to this PR https://github.com/SAP/wdio-qmate-service/pull/30 so we don't mix up the remove legacy with workflow settings. Also I have added the missing jobs there.

cvakiitho (dismissed):

PR #36, opened 2022-08-15
change to normal util tests

DIFSRIP:
- enables data privacy tests
- requires working secret keys


z164:
@DIFSRIP Removed binding to repo and reverted the keys to the OG ones. For the PW i've added env variable and the default value. Looks fine ?

z164:
Duplicating the approach to the qmate as well

src/scripts/hooks/utils/decryption.ts
```diff
@@ -88,21 +88,17 @@ class Decryption {
     }
   }
 
-  private _decryptDataWithRepoName(data: Buffer) {
```
DIFSRIP, line 91: same here as in qmate repo

DIFSRIP (commented):
Please see my comments

.github/workflows/node.js.yml
```diff
@@ -145,14 +145,11 @@ jobs:
...
           node-version: ${{ matrix.node-version }}
-          cache: 'npm'
       - name: run tests
         env: 
           QMATE_PRIVATE_KEY: ${{ secrets.QMATE_PRIVATE_KEY }}
         run: |
-          sudo apt-get update
-          sudo apt-get install -y --only-upgrade openssl
-          sudo apt-get install google-chrome-stable
+          echo "$QMATE_PRIVATE_KEY" > private.key
```
DIFSRIP, line 152: @cvakiitho will this persist in pipeline and will it be accessible from outside? Since we remove the key during qmate init, we need to make sure that we do not have any public copies of it

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sap_oss/wdio-qmate-service",
-  "version": "1.0.1",
+  "version": "1.1.1",
```
DIFSRIP, line 3: its 1.0.2

DIFSRIP: please update the lock file with `npm install`

DIFSRIP (commented):
please see my comments

.github/workflows/node.js.yml
z164: removed since it does not affect anything. key is being picked up from env correcly.

cvakiitho: The VMs are removed after pipeline is finished, so it won't metter - but yeah- this was only a hack to see it's correctly passed, and it made some encoding issues dissapear in jenkinses.

cvakiitho:
Just thinking, I see it works fine, that's awesome, but couldn't we just use a encryption type that doesn't have line breaks?
I guess that would be a breaking change, where teams would have to re-generated their encrypted data right.

DIFSRIP:
> Just thinking, I see it works fine, that's awesome, but couldn't we just use a encryption type that doesn't have line breaks? I guess that would be a breaking change, where teams would have to re-generated their encrypted data right.

I guess if we want to use RSA there is no way around this

ickitc (approved):

ickitc (approved):

PR #47, opened 2022-08-23
add option to specify authentication type for odata service

heraldantony:
For "communication" users SAML login does not work.
Add authentication type argument to `service.odata.init` function, to override the default SAML authentication.


DIFSRIP (approved):
looks good

PR #50, opened 2022-08-29
fix doc generation script for windows

heraldantony:
`glob` is having issues with back slashes in path name on windows systems, so doc generation fails.
Replace back slashes with forward slashes. 
Should not affect anything on Unix systems.


DIFSRIP (dismissed):
looks good

DIFSRIP (approved):
looks good

PR #52, opened 2022-09-05
fix spec for util.browser.switchToNewWindow

heraldantony:
spec fails because the demo app uses a different theme


DIFSRIP (approved):
looks good

PR #75, opened 2022-11-13
feat: parallel login

z164:


DIFSRIP (approved):
looks good to me!

@cvakiitho any concerns about running Promise.any in pipelines because of the node version?

PR #79, opened 2022-11-22
circular reference error while getting ui5 aggregation property

heraldantony:
For some properties `ui5.control.getAggregationProperty` fails with "javascript error: circular reference"
E.g. `sap.m.Table` has columns property, each column has reference to the parent table, and the parent has columns, leading to circular references.
Code available here https://github.com/douglascrockford/JSON-js/blob/master/cycle.js is used to remove object references, so that webdriver can serialize the object to JSON string and return it.
This serialized object is then restored.

### DIFF
--- a/src/scripts/hooks/utils/cycle.js
+++ b/src/scripts/hooks/utils/cycle.js
@@ -1,0 +1,179 @@
+/*
+    cycle.js
+    2021-05-31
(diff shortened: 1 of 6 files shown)


DIFSRIP (approved):
looks good

PR #82, opened 2022-11-28
nonUi5.userInteraction.mouseOverElement does not work for elements within iframe

heraldantony:
Add tests for nonUi5.userInteraction.mouseOverElement with iframes
1. Main window doesn't have any content, so iframe is at position (0,0) - mouseOverElement works in this case
2. Main window has content, so the included iframe is at a non-zero position - mouseOverElement does not work


Templada:
@heraldantony can we close this one as well?

heraldantony:
@Templada, we can close this issue. Was reported by just one team using Sidekick extension.

PR #86, opened 2022-12-01
reduce amount of data returned by ui5.control.getAggregationProperty

heraldantony:
The data returned after decycling cyclic references is too big, and the call fails.
Remove unnecessary properties "oParent", "oPropagatedProperties" and "undefined" from returned data


src/reuse/modules/ui5/types/ui5.types.ts
```diff
@@ -0,0 +1,9 @@
+export type Ui5Selector = {
+  elementProperties: any
+};
+export type Ui5SelectorOptions = {
```
Templada, line 4: @DIFSRIP should we rename SelectorOptions?

src/scripts/hooks/utils/cycle.js
```diff
@@ -104,10 +102,16 @@ module.exports = {
 
           nu = {};
```
Templada, line 103: what stands 'nu' for?

src/scripts/hooks/utils/locatorCommands.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     return lib.uiControlExecuteLocator(ui5Selector, null, timeout, rootElement);
   },
 
-  getUI5Property: async function (propName, selectorOrElement) {
+  getUI5Property: async function (propName, element) {
```
Templada, line 14: why did we change to element only here? Is it just bc we already got the element at this level?

DIFSRIP (commented):
please see my comments

src/reuse/modules/ui5/types/ui5.types.ts
Templada: agree. @heraldantony Can you change to "Ui5SelectorWithOptions" please?

src/reuse/modules/ui5/types/ui5.types.ts
```diff
@@ -0,0 +1,9 @@
+export type Ui5Selector = {
+  elementProperties: any
```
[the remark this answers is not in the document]
Templada: +1

Templada (changes requested):
@heraldantony please see the comments. A few naming changes are requested.

src/scripts/hooks/utils/locatorCommands.js
heraldantony: will fix that, that code shouldn't have been changed

src/reuse/modules/ui5/types/ui5.types.ts
heraldantony: yes, changed to Ui5SelectorWithOptions

src/scripts/hooks/utils/cycle.js
heraldantony: don't know why :)  Code comes from here - https://github.com/douglascrockford/JSON-js/blob/master/cycle.js, and this is also referred to in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value

src/reuse/modules/ui5/types/ui5.types.ts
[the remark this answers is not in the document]
heraldantony: Added optional values, and fixed types.

DIFSRIP (approved):
lgtm

Templada (approved):

PR #94, opened 2022-12-08
clear works only if user tabs out of field

heraldantony:
ui5.userInteraction.clear - this clears the field on screen, but the element property value does not change, so getValue returns the old value.
ui5.userInteraction.clearAndRetry - does not verify the property value

selectAll spec is included here, since the verification step requires pressing backspace, and Enter, before checking if value is cleared.


Templada:
@heraldantony can we close this. Quite old this one

heraldantony:
@Templada - yes, old issue, closing.

PR #97, opened 2022-12-14
add tests for new util.data and util.component reuse functions

heraldantony:
Add tests for 
util.data.writeDataToFile
util.data.readDataFromFile
util.component.loadEntryPoint
util.component.storeEntryPoint


DIFSRIP (approved):
Great! Looks good!

PR #98, opened 2022-12-19
Verbose mode

z164:


z164:
@DIFSRIP basically we do not have a unified .prettierrc.json (or analogue). want me to implement one, that i used for this kind of formatting? tried not to change much, but still if you want smth to be different, you can ping :D

PR #99, opened 2022-12-23
add ui5.table tests for latest ui5 version

heraldantony:
Table column headers in newer ui5 versions don't have the tooltip. We don't use tooltip value in the selectors in `ui5.table`, so `ui5.table.sortColumnAscending` and `ui5.table.sortColumnDescending` work fine with newer ui5 versions.

Current tests use ui5 version 1.99.0, and have column header selectors with tooltip value.
Since newer ui5 versions don't have tooltips for column headers, added tests for new ui5, with proper selectors.

### DIFF
--- a/test/reuse/ui5/table/clickSettingsButton.spec.js
+++ b/test/reuse/ui5/table/clickSettingsButton.spec.js
@@ -56,3 +56,59 @@ describe("table - clickSettingsButton - smartTable with tableSelector", function
   });
 
 });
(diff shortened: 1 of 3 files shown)


heraldantony:
Fix for sorting is in PR - https://github.com/SAP/wdio-qmate-service/pull/100
Don't require these additional tests for ui5 > 1.99, since they don't add anything.

PR #104, opened 2023-01-10
add loginNetWeaver function

Templada:


src/reuse/modules/ui5/session.ts
```diff
@@ -132,6 +132,25 @@ export class Session {
...
+  /**
+ * @function loginNetWeaver
+ * @memberOf ui5.session
+ * @description Login with sap netweaver for WebGUI form and specific username and password.
+ * @param {String} username - The username.
+ * @param {String} [password] - The password.
+ * @param {Boolean} [pressContinue=true] - Specifies if the function will press continue if applicable.
+ * @example await ui5.session.loginNetWeaver("john", "abc123!");
+ */
+  async loginNetWeaver(username: string, password?: string, pressContinue = true) {
```
Templada, line 144: thinking of moving it into nonUi5... wdyt? @DIFSRIP

DIFSRIP: I would move it to common maybe? Maybe the other logins as well at some point of time?

DIFSRIP (commented):
looks good to me, maybe we can ad a short test for it. Just for local execution and / or a static one as for the others.

Templada:
> looks good to me, maybe we can ad a short test for it. Just for local execution and / or a static one as for the others.

fine with me, then i just move the loginNetWeaver to common and create a new session module

DIFSRIP (approved):

PR #112, opened 2023-02-07
raw header response for POST & GET

Templada:


src/reuse/modules/service/odata.ts
```diff
@@ -197,14 +203,17 @@ export class OData {
...
    * let res = await service.odata.post(srv, "A_PurchaseOrder", payload);
    */
-  async post (srv: any, entitySet: string, payload: any) {
+  async post(srv: any, entitySet: string, payload: any, raw: boolean = false) {
     const entity = srv[entitySet];
     if (!entity) {
       throw new Error(`No entity set '${entitySet}' available in service`);
     }
-    return entity.post(payload);
+    if (raw) {
```
DIFSRIP, line 211: better check for raw === true, otherwise it could be anything except for null, undefined and false
```suggestion
    if (raw === true) {
```

DIFSRIP (changes requested):
please see comment

src/reuse/modules/service/odata.ts
```diff
@@ -92,19 +92,24 @@ export class OData {
...
    * const res = await service.odata.get(srv, "A_PurchaseOrder", keys);
    */
-  async get (srv: any, entitySet: string, keys: any): Promise<any> {
+  async get(srv: any, entitySet: string, keys: any, raw: boolean = false): Promise<any> {
     const entity = srv[entitySet];
     if (!entity) {
       throw new Error(`No entity set '${entitySet}' available in service`);
     }
-    return entity.get(keys);
+    if (raw) {
```
Templada, line 108:
```suggestion
    if (raw === true) {
```

DIFSRIP (approved):

PR #114, opened 2023-02-15
Custom runner commands

DIFSRIP:
Adds custom it (cit -> component it) for Qmate Proxy which is skipped in REPLAY mode.
Can be used for login, closePopups and other functions which should only be executed in RECORD.


heraldantony (approved):
looks good

PR #119, opened 2023-03-09
add new date format: mmm d, yyyy

DIFSRIP:
solves: https://github.tools.sap/sProcurement/qmate/issues/318

### DIFF
--- a/src/reuse/modules/util/formatter.ts
+++ b/src/reuse/modules/util/formatter.ts
@@ -141,7 +141,11 @@ export class Formatter {
    * // returns "Apr 03, 2022"
    */
   formatDate(date: Date, format: DateFormatsType, locale = "en-US"): string | Date {
+    if (format) {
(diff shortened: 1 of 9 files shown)


src/reuse/modules/util/formatter.ts
```diff
@@ -162,7 +162,7 @@ export class Formatter {
       hour = `0${hour}`;
     }
 
-    if (dd < 10) {
+    if (dd < 10 && format !== DateFormats.MONTH_DAY_YEAR_COMMA_SHORT) {
```
heraldantony, line 165: Should lowercase format before this check, the test fails otherwise.

DIFSRIP: right, will do that

heraldantony (dismissed):
looks good

heraldantony (approved):
looks good

PR #131, opened 2023-04-13
Fix the issue related to selectBox when option is empty string

ihappyk:
Fix for https://github.tools.sap/sProcurement/qmate/issues/328

### DIFF
--- a/src/reuse/modules/ui5/userInteraction.ts
+++ b/src/reuse/modules/ui5/userInteraction.ts
@@ -427,7 +427,7 @@ export class UserInteraction {
   async selectBox(selector: any, value: string, index = 0) {
     const vl = this.vlf.initLog(this.selectBox);
     await this.clickSelectArrow(selector, index);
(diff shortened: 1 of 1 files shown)


heraldantony (dismissed):
looks good

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -427,7 +427,7 @@ export class UserInteraction {
   async selectBox(selector: any, value: string, index = 0) {
     const vl = this.vlf.initLog(this.selectBox);
     await this.clickSelectArrow(selector, index);
-    if (value) {
+    if (value !== undefined) {
```
DIFSRIP, line 430:
```suggestion
    if (value !== undefined && value !== null) {
```

DIFSRIP (changes requested):
please see my suggestion

DIFSRIP (approved):
looks good

PR #133, opened 2023-04-13
Fix clearAndFill for checkboxes

PhyNerd:


src/reuse/modules/nonUi5/userInteraction.ts
```diff
@@ -237,7 +237,7 @@ export class UserInteraction {
   async clearAndFill(element: Element, value: string) {
     const vl = this.vlf.initLog(this.clearAndFill);
     //arg. 'value' needs to be checked in case of numeric values. E.g.: 0 or 1 will be handled as boolean value in if.
-    if (!element || !(typeof value === "number" || typeof value === "string")) {
+    if (!element || !(typeof value === "number" || typeof value === "string" || typeof value === "boolean")) {
```
DIFSRIP, line 240: same change for fill?

PR #134, opened 2023-04-13
add check/uncheck

DIFSRIP:
- Add requests function to check/uncheck select boxes (ui5 / nonui5)
- Solves issue with limited type for fill.

### DIFF
--- a/src/reuse/modules/common/userInteraction.ts
+++ b/src/reuse/modules/common/userInteraction.ts
@@ -15,20 +15,21 @@ export class UserInteraction {
    * @function fillActive
    * @memberOf common.userInteraction
    * @description Fills the active input with the given value.
-   * @param {String} value - The value with witch the input should be filled.
+   * @param {String | Number | Boolean} value - The value to enter.
    * @example await common.userInteraction.fillActive("My Value");
    */
-  async fillActive(value: string) {
+  async fillActive(value: string | number | boolean) {
     const vlf = this.vlf.initLog(this.fillActive);
-    if (!(typeof value === "number" || typeof value === "string")) {
-      throw new Error("Function 'fillActive' failed: Please provide a value(datatype - number/string) as argument.");
-    } else {
+
+    if (typeof value === "number" || typeof value === "string" || typeof value === "string") {
       try {
         const elem = await $(await browser.getActiveElement());
         await elem.addValue(value);
       } catch (error) {
         throw new Error(`Function 'fillActive' failed: ${error}`);
       }
+    } else {
+      throw new Error("Function 'fillActive' failed: Please provide a value(datatype - number/string) as argument.");
     }
   }
 
@@ -36,7 +37,7 @@ export class UserInteraction {
    * @function fillActiveAndRetry
    * @memberOf common.userInteraction
    * @description Enters the given value to the active input field and retries the action in case it fails.
-   * @param {String} value - The value with witch the input should be filled.
+   * @param {String} value - The value to enter.
    * @param {Number} [retries=3] - The number of retries, can be set in config for all functions under params stepsRetries.
    * @param {Number} [interval=5000] - The delay between the retries (ms). Can be set in config for all functions under params.stepRetriesIntervals.
    * @example await common.userInteraction.fillActiveAndRetry("My Value");
--- a/src/reuse/modules/common/userInteraction.ts
+++ b/src/reuse/modules/common/userInteraction.ts
@@ -15,14 +15,16 @@ export class UserInteraction {
    * @function fillActive
    * @memberOf common.userInteraction
    * @description Fills the active input with the given value.
-   * @param {String | Number | Boolean} value - The value to enter.
+   * @param {String | Number} value - The value to enter.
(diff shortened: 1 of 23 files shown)


src/reuse/modules/nonUi5/element.ts
```diff
@@ -423,6 +423,20 @@ export class ElementModule {
...
+   * @memberOf nonUi5.element
+   * @description Returns a boolean if the element (e.g. checkbox) is selected.
+   * @param {Object} elem - The element.
+   * @returns {boolean}
+   * @example const elem = await nonUi5.element.getById("elem01");
+   * const isSelected = await nonUi5.element.isSelected(elem);
+   */
+  async isSelected(elem: Element): Promise<boolean> {
+    const vl = this.vlf.initLog(this.isSelected);
+    return elem.isSelected();
```
cvakiitho, line 437: Why are we wrapping wdio element method, for better visibility to teams?

src/reuse/modules/common/userInteraction.ts
```diff
@@ -15,28 +15,29 @@ export class UserInteraction {
...
+
+    if (typeof value === "number" || typeof value === "string" || typeof value === "string") {
       try {
         const elem = await $(await browser.getActiveElement());
         await elem.addValue(value);
       } catch (error) {
         throw new Error(`Function 'fillActive' failed: ${error}`);
       }
+    } else {
+      throw new Error("Function 'fillActive' failed: Please provide a value(datatype - number/string) as argument.");
```
cvakiitho, line 32: I'd expect TS to catch this in compiler, but this is probably a fallback for JS tests right?

src/reuse/modules/common/userInteraction.ts
```diff
@@ -15,28 +15,29 @@ export class UserInteraction {
...
    * @example await common.userInteraction.fillActive("My Value");
    */
-  async fillActive(value: string) {
+  async fillActive(value: string | number | boolean) {
     const vlf = this.vlf.initLog(this.fillActive);
-    if (!(typeof value === "number" || typeof value === "string")) {
-      throw new Error("Function 'fillActive' failed: Please provide a value(datatype - number/string) as argument.");
-    } else {
+
+    if (typeof value === "number" || typeof value === "string" || typeof value === "string") {
```
cvakiitho, line 24: duplicate string here - but as previous comment - it probably can only be number/string?

src/reuse/modules/nonUi5/element.ts
DIFSRIP: - we wrap everything, to be able to keep compatibility (same as moving from vyper to qmate, or if wdio does changes)
- to keep consistent usage (we have function where we pass the elem as argument, wdio has functions on elements)
- to provide flexibility and additional checks (in this case only the vl)
- transparency through Reuse API

src/reuse/modules/common/userInteraction.ts
DIFSRIP: sure, most tests are written in js, so we need to provide proper error handling.

src/reuse/modules/common/userInteraction.ts
```diff
@@ -15,28 +15,29 @@ export class UserInteraction {
    * @function fillActive
    * @memberOf common.userInteraction
    * @description Fills the active input with the given value.
-   * @param {String} value - The value with witch the input should be filled.
+   * @param {String | Number | Boolean} value - The value to enter.
    * @example await common.userInteraction.fillActive("My Value");
    */
-  async fillActive(value: string) {
+  async fillActive(value: string | number | boolean) {
```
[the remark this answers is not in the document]
DIFSRIP: I know, strangely SIV team was using this before with boolean value to check/uncheck checkboxes. @PhyNerd requested to support this but I am not sure if and how it works.

src/reuse/modules/common/userInteraction.ts
DIFSRIP: thanks will remove the duplicate check

src/reuse/modules/nonUi5/userInteraction.ts
```diff
@@ -219,35 +280,39 @@ export class UserInteraction {
...
    * @memberOf nonUi5.userInteraction
    * @description Clears and fills the passed input element.
    * @param {Object} element - The element.
-   * @param {String} value - The value to be filled in.
+   * @param {String | Number} value - The value to enter in.
    * @example const elem = await nonUi5.element.getById("input01");
    * await nonUi5.userInteraction.clearAndFill(elem, "Service 01");
    */
-  async clearAndFill(element: Element, value: string) {
+  async clearAndFill(element: Element, value: string | string) {
```
heraldantony, line 303: typo - string | number

heraldantony (approved):
looks good

src/reuse/modules/nonUi5/userInteraction.ts
DIFSRIP: done

heraldantony (approved):
looks good

PR #153, opened 2023-05-15
remove deepmerge

Varun-Hegde:
Removing deepmerge of browser.options and config, as config will have all the command line args itself.

Tested with appCov profile and using command line args, works fine

 ```sudo apt-get install google-chrome-stable``` installs the 113.0.0 version. So have updated the dependency


heraldantony (approved):
looks good

PR #156, opened 2023-05-24
Designed utility for excel file read and recursively find file path for given any file

Ananthreddy3004:
util.file.getExcelData('/Users/', 'test.xlx') :- this reads the entire excel file based on provided sheet number and return the excel content in the format of json by default but it can also be csv, txt based on our requirement excel data can be converted 

util.file.getAbsoluteFilePath('/Users','test.xls') :- It return the absolute file path for given file which requires the parameters root directory and file to be searched


test/reuse/util/file/getExcelData.spec.js
```diff
@@ -0,0 +1,45 @@
+const path = require("path");
+
+describe("file - getExcelData - Retrieve xls file data", function () {
```
DIFSRIP, line 3: lets add some more tests, testing the different output formats

test/reuse/util/file/getExcelData.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  let value;
+
+  it("Execution", async function () {
+    const xlsPath = path.resolve(__dirname, "./testFiles");
+    value = await util.file.getExcelData(xlsPath, "test.xls", 1);
+  });
+
+  it("Verification", async function () {
+    common.assertion.expectDefined(value);
```
DIFSRIP, line 12: instead of the generic assertion, you can check for the exact output or assert parts of the output

DIFSRIP (changes requested):
Please see my comments

test/reuse/util/file/getExcelData.spec.js
Ananthreddy3004: @DIFSRIP I have added other output formats in test

test/reuse/util/file/getExcelData.spec.js
Ananthreddy3004: I have added specific assertion to test instead of generic

DIFSRIP (approved):

PR #158, opened 2023-06-01
add file upload reuse function for WebGUI apps

heraldantony:
Add reuse function for file upload in WebGUI apps.

### DIFF
--- a/src/reuse/modules/util/file.ts
+++ b/src/reuse/modules/util/file.ts
@@ -45,6 +45,43 @@ export class File {
       throw new Error(`Function 'upload' failed': ${error}`);
     }
   }
+  
+  /**
+   * @function uploadWebGui
+   * @memberOf util.file
+   * @description Uploads all the file/s by the paths given in the Array for SAP WebGUI apps.
+   * @param {String[]} files - Array with path/s of file/s to be uploaded.
+   * @param {String | Object} selector - Custom selector of the input element
+   * @example await util.file.uploadWebGui(["path/to/text1.txt"], "INPUT[title='External file name']");
+   */
+  async uploadWebGui(files: Array<string>, selector: any) {
+    const vl = this.vlf.initLog(this.uploadWebGui);
+    try {
+      if (typeof selector === "string") {
+        const elem = await nonUi5.element.getByCss(selector);
+        await nonUi5.userInteraction.click(elem);
+      } else if (typeof selector === "object") {
+        await ui5.userInteraction.click(selector);
+      }
+      await common.userInteraction.pressF4();
+      const okButton = await nonUi5.element.getByCss("DIV[id='UpDownDialogChoose']")
+      await nonUi5.assertion.expectToBeVisible(okButton);
+      const fileInput = await nonUi5.element.getByXPath(".//input[@id='webgui_filebrowser_file_upload'][@type='file']", 0, 30000, true);
+      let remoteFiles = "";
+      for (const file of files) {
+        const filePath = this.path.resolve(file);
+        vl.log(`Uploading file with a path ${filePath}`);
+        const remoteFilePath = await browser.uploadFile(filePath);
+        if (remoteFiles) {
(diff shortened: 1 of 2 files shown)


Templada (dismissed):
+1

src/reuse/modules/util/file.ts
```diff
@@ -45,6 +45,43 @@ export class File {
...
+   * @example await util.file.uploadWebGui(["path/to/text1.txt"], "INPUT[title='External file name']");
+   */
+  async uploadWebGui(files: Array<string>, selector: any) {
+    const vl = this.vlf.initLog(this.uploadWebGui);
+    try {
+      if (typeof selector === "string") {
+        const elem = await nonUi5.element.getByCss(selector);
+        await nonUi5.userInteraction.click(elem);
+      } else if (typeof selector === "object") {
+        await ui5.userInteraction.click(selector);
```
DIFSRIP, line 64: do we need to support the ui5 version in the webgui reuse?

src/reuse/modules/util/file.ts
```diff
@@ -45,6 +45,43 @@ export class File {
...
+      }
+      await common.userInteraction.pressF4();
+      const okButton = await nonUi5.element.getByCss("DIV[id='UpDownDialogChoose']")
+      await nonUi5.assertion.expectToBeVisible(okButton);
+      const fileInput = await nonUi5.element.getByXPath(".//input[@id='webgui_filebrowser_file_upload'][@type='file']", 0, 30000, true);
+      let remoteFiles = "";
+      for (const file of files) {
+        const filePath = this.path.resolve(file);
+        vl.log(`Uploading file with a path ${filePath}`);
```
DIFSRIP, line 73:
```suggestion
        vl.log(`Uploading file with path ${filePath}`);
```

src/reuse/modules/util/file.ts
```diff
@@ -45,6 +45,43 @@ export class File {
...
+      if (typeof selector === "string") {
+        const elem = await nonUi5.element.getByCss(selector);
+        await nonUi5.userInteraction.click(elem);
+      } else if (typeof selector === "object") {
+        await ui5.userInteraction.click(selector);
+      }
+      await common.userInteraction.pressF4();
+      const okButton = await nonUi5.element.getByCss("DIV[id='UpDownDialogChoose']")
+      await nonUi5.assertion.expectToBeVisible(okButton);
+      const fileInput = await nonUi5.element.getByXPath(".//input[@id='webgui_filebrowser_file_upload'][@type='file']", 0, 30000, true);
```
DIFSRIP, line 69: please use getByCss, internally the same is called anyways: https://github.com/SAP/wdio-qmate-service/blob/main/src/reuse/modules/nonUi5/element.ts#L269

```suggestion
      const fileInput = await nonUi5.element.getByCss(".//input[@id='webgui_filebrowser_file_upload'][@type='file']", 0, 30000, true);
```

DIFSRIP (changes requested):
please see my comments

src/reuse/modules/util/file.ts
Templada: agree

src/reuse/modules/util/file.ts
Templada: also valid

src/reuse/modules/util/file.ts
Templada: what do you mean by that?

heraldantony: fixed, changed selector type to string. The input element won't be a ui5 element.

src/reuse/modules/util/file.ts
heraldantony: fixed :)

src/reuse/modules/util/file.ts
heraldantony: changed to getByCss

DIFSRIP (approved):
looks good

PR #164, opened 2023-06-15
add user module

Templada:


src/reuse/modules/util/userSettings.ts
```diff
@@ -0,0 +1,94 @@
...
+export class UserSettings {
+  private vlf = new VerboseLoggerFactory("util", "user");
+  private _srvInstance = null;
+
+  private async _initForUserSetting(user: string, password: string): Promise<any> {
+    const vl = this.vlf.initLog(this._initForUserSetting);
+    //TODO: Add vl.log msg
+    const params = browser.config.params;
+    if (params?.systemUrl) {
+      this._srvInstance = await service.odata.init(`${params.systemUrl}/sap/opu/odata/UI2/INTEROP`, user, password);
```
Templada, line 19: browser.config.params.odataUrl can not be taken since the default for odataUrl is
`/sap/opu/odata/sap` and the `INTEROP`service url is `/sap/opu/odata/UI2/INTEROP/` so the automatic concat of `/sap/opu/odata/sap/UI2/INTEROP`would lead to an incorrect URL

test/reuse/util/userSettings/public.key
```diff
@@ -0,0 +1,8 @@
+-----BEGIN RSA PUBLIC KEY-----
```
Templada, line 1: @DIFSRIP | @cvakiitho : do we need to set/overwrite the priv key in the gh-actions

test/reuse/util/userSettings/test.user.conf.js
```diff
@@ -0,0 +1,20 @@
+const path = require("path");
+const merge = require("deepmerge");
+const qmateConfiguration = require("../../../helper/configurations/chrome.headless.conf");
+const data = require("./data/data.local.json");
+exports.config = merge(qmateConfiguration.config, {
+  maxInstances: 1,
+  params: {
+    systemUrl: data.systemUrl,
```
Templada, line 8: @DIFSRIP: Lets dicsuss the data handling

src/reuse/modules/util/formatter.ts
```diff
@@ -143,7 +143,7 @@ export class Formatter {
    * const formattedDate = util.formatter.formatDate(date, "mmm dd, yyyy");
    * // returns "Apr 03, 2022"
    */
-  formatDate(date: Date, format: DateFormatsType, locale = "en-US"): string | Date {
+  formatDate(date: Date, format: DateFormatsType, locale = "en-US"): string | Date { //TODO: suggestion to add process.env.USER_SETTINGS_DATE_FORMAT as a default value for format
```
Templada, line 146: let's discuss this (comment)

src/reuse/modules/common/date.ts
```diff
@@ -23,7 +23,7 @@ export class DateModule {
    * @returns {String} The date in the given format.
    * @example const date = await common.date.getToday("mm/dd/yyyy");
    */
-  getToday(format: DateFormatsType = DateFormats.OBJECT): Date | string {
+  getToday(format: DateFormatsType = process.env.USER_SETTINGS_DATE_FORMAT as DateFormats.OBJECT): Date | string {
```
vlad9719, line 26: Is it possible that process.env.USER_SETTINGS_DATE_FORMAT is not defined here, for instance if no login happens? If so, I would add a fallback default option, here and in the other places in this module
```suggestion
  getToday(format: DateFormatsType = process.env.USER_SETTINGS_DATE_FORMAT ?  process.env.USER_SETTINGS_DATE_FORMAT as DateFormats.OBJECT : DateFormats.OBJECT): Date | string {
```

src/reuse/modules/util/constants/formatter.constants.ts
```diff
@@ -8,5 +8,17 @@ export enum DateFormats {
   MONTH_DAY_YEAR_COMMA = "mmm dd, yyyy",
   MONTH_DAY_YEAR_COMMA_SHORT = "mmm d, yyyy",
   DATETIME = "datetime",
-  OBJECT = "object"
-}
+  OBJECT = "object",
+  GREGORIAN_DOT = "dd.mm.yyyy (Gregorian Date)",
```
vlad9719, line 12: Is it necessary to add those brackets with comment in the end, here and in the constants below?

test/reuse/util/userSettings/test.user.conf.js
```diff
@@ -0,0 +1,20 @@
+const path = require("path");
+const merge = require("deepmerge");
+const qmateConfiguration = require("../../../helper/configurations/chrome.headless.conf");
+exports.config = merge(qmateConfiguration.config, {
+  params: {
+    import: {
+      data: "./data/"
+    },
+    // applyUserSettingsForS4: true
```
vlad9719, line 9: Do you need this param in the config? If not, consider removing it

src/reuse/modules/util/formatter.ts
```diff
@@ -156,64 +156,44 @@ export class Formatter {
...
+
+    sec = sec.toString().padStart(2, "0");
+    min = min.toString().padStart(2, "0");
+    hour = hour.toString().padStart(2, "0");
+    dd = (dd < 10 && format !== DateFormats.MONTH_DAY_YEAR_COMMA_SHORT) ? dd.toString().padStart(2, "0") : dd;  
+    mm = mm.toString().padStart(2, "0");
+    yy = yy.toString().padStart(2, "0");
+
+
+    const formatMappings = {
```
vlad9719, line 169: I like this constant, but I would keep it outside of the class, just to make the function a bit shorter and easier to read?

src/reuse/modules/util/formatter.ts
```diff
@@ -156,64 +156,44 @@ export class Formatter {
...
-    }
-
-    if (dd < 10 && format !== DateFormats.MONTH_DAY_YEAR_COMMA_SHORT) {
-      dd = `0${dd}`;
+    let yy: number | string = yyyy % 100;
+
+    sec = sec.toString().padStart(2, "0");
+    min = min.toString().padStart(2, "0");
+    hour = hour.toString().padStart(2, "0");
+    dd = (dd < 10 && format !== DateFormats.MONTH_DAY_YEAR_COMMA_SHORT) ? dd.toString().padStart(2, "0") : dd;
```
vlad9719, line 164: I would move the part of the expression before '?' into a separate local boolean function, `shouldPadDay`. Thus readability could be improved

vlad9719 (changes requested):
Thank you for your work!

Looks good overall, but I've added a couple of suggestions, and also questions, where I did not understood the code. Please have a look at those

src/reuse/modules/util/userSettings.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (!this._srvInstance) {
+      await this._initForUserSetting(user, password);
+    }
+    const res = await service.odata.get(this._srvInstance, "UserProfileProperties", { id: "NUMBER_FORMAT", shellType: "FLP" });
+    const resUserData = await service.odata.get(this._srvInstance, "UserProfilePropertyValues", { id: "NUMBER_FORMAT", shellType: "FLP", value: res.value });
+    process.env.USER_SETTINGS_NUMBER_FORMAT = resUserData.description;
+    util.console.info(`Number Format: ${process.env.USER_SETTINGS_NUMBER_FORMAT} was set.`);
+  }
+
+  public async apply(user: string, password: string, isS4: boolean) {
```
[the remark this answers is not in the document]
Templada: will be browser.config.params.applyS4UserSettings or so then in the end

test/reuse/util/userSettings/setDateFormatFromUserSettings.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution & Verification: Set User Date", async function () {
+      await util.userSettings.setDateFormatFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
+      const date = await common.date.getToday("yyyy/mm/dd");
+      const userDate = await common.date.getToday(process.env.USER_SETTINGS_DATE_FORMAT);
+      //Ensure returned date matches 10 digits
+      common.assertion.expectEqual(date.length, 10);
```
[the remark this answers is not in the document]
Templada: got it, i can replace it with a meaningful constant - it is 10 digits because of the given date formated passed in to the getToday (yyyy/mm/dd), so it should return sth like 2024/08/02 which is 10 chars long.

src/reuse/modules/common/date.ts
Templada: i think that should already be the case, but would be more clear with the ternary expression

src/reuse/modules/util/constants/formatter.constants.ts
Templada: before it was, as the response of the odata request is with the brackets, since i cut it now before brackets with replace(), we can probably also delete it from the formatter.constants.ts

src/reuse/modules/util/formatter.ts
Templada: actually i didnt changed that at all, think it is showing as new code as it got slightly moved?

src/reuse/modules/util/userSettings.ts
```diff
@@ -0,0 +1,93 @@
...
+    } else {
+      throw new Error("System URL is missing in the config file.");
+    }
+  }
+
+  public async setLanguageFromUserSettings(user: string, password: string) {
+    const vl = this.vlf.initLog(this.setLanguageFromUserSettings);
+    if (!this._srvInstance) {
+      await this._initForUserSetting(user, password);
+    }
```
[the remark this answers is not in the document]
Templada: moved it to the init and called in every method the init

test/reuse/util/userSettings/setDateFormatFromUserSettings.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution & Verification: Set User Date", async function () {
+      await util.userSettings.setDateFormatFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
+      const date = await common.date.getToday("yyyy/mm/dd");
+      const userDate = await common.date.getToday(process.env.USER_SETTINGS_DATE_FORMAT);
+      //Ensure returned date matches 10 digits
+      common.assertion.expectEqual(date.length, DATE_FORMAT_LENGTH);
```
Templada, line 17: expect type of enum dateformat object instead of length

src/reuse/modules/util/userSettings.ts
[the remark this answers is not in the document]
vlad9719: have you pushed that change? I still see it as before

src/reuse/modules/util/userSettings.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (browser.config.params?.applyUserSettingsForS4) {
+      try {
+        await this._initForUserSetting(user, password);
+        await this.setDateFormatFromUserSettings(user, password);
+        await this.setLanguageFromUserSettings(user, password);
+        await this.setNumberFormatFromUserSettings(user, password);
+        await this.setTimeFormatFromUserSettings(user, password);
+        await this.setTimeZoneFromUserSettings(user, password);
+      } catch (error) {
+      vl.log(`Function: 'setUserSettingsForS4' failed: Unable to set the UserSettings: ${error}`);
```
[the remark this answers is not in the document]
vlad9719: so basically this error would be thrown if someone sets userSettings in the browser.config.params, but they could not be set? Then I would suggest write the message just as `Failed to set the UserSettings for s4: ${error}`

src/reuse/modules/util/formatter.ts
DIFSRIP: but the code is very different, can you please double check

src/reuse/modules/util/formatter.ts
DIFSRIP: can we move it out nevertheless as Vlad suggested?

src/reuse/modules/util/userSettings.ts
```diff
@@ -0,0 +1,129 @@
...
+
+import { VerboseLoggerFactory } from "../../helper/verboseLogger";
+
+/**
+ * @class userSettings
+ * @memberof util
+ */
+
+export class UserSettings {
+  private vlf = new VerboseLoggerFactory("util", "user");
```
DIFSRIP, line 11:
```suggestion
  private vlf = new VerboseLoggerFactory("util", "userSettings");
```

src/reuse/modules/util/userSettings.ts
```diff
@@ -0,0 +1,129 @@
...
+      } else {
+        throw new Error("System URL is missing in the config file.");
+      }
+    }
+  }
+
+  public async setLanguageFromUserSettings(user: string, password: string) {
+    const vl = this.vlf.initLog(this.setLanguageFromUserSettings);
+    await this.initS4UserSettingService(user, password);
+    const res = await service.odata.get(this._srvInstance, "UserProfileProperties", { id: "LANGUAGE", shellType: "FLP" });
```
DIFSRIP, line 37: I would try catch all the API calls as you did for the init.

src/reuse/modules/util/userSettings.ts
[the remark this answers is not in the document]
DIFSRIP: in general, the logger should not be used for errors. Errors will not be shown in regular mode like that. It is used to log additional info on the function behaviour if executed in verbosity mode. You should throw a regular error

test/reuse/util/userSettings/getDateFormatFromUserSettings.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution & Verification: Set User Date", async function () {
+      const userDateFormat = await util.userSettings.getDateFormatFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
+      const date = await common.date.getToday("yyyy/mm/dd");
+      const userDate = await common.date.getToday(userDateFormat);
+      //Ensure returned date matches 10 digits
+      common.assertion.expectEqual(date.length, DATE_FORMAT_LENGTH);
```
DIFSRIP, line 17: instead of checking for the length (which might not be working for specific other configs), I would check if both dates are of the same type (DateFormatsType) since the userDateFormat could be different from time to time while the "yyyy/mm/dd" is static

test/reuse/util/userSettings/getTimeFormatFromUserSettings.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe("userSettings", function () {
+  describe("getTimeFormatFromUserSettings.spec", function () {
+
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution", async function () {
+      await util.userSettings.getTimeFormatFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
```
DIFSRIP, line 12: no assertion here?

test/reuse/util/userSettings/getTimeZoneFromUserSettings.spec.js
```diff
@@ -0,0 +1,15 @@
...
+
+describe("userSettings", function () {
+  describe("getTimeZoneFromUserSettings.spec", function () {
+
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution", async function () {
+      await util.userSettings.getTimeZoneFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
```
DIFSRIP, line 11: no assertion here?

test/reuse/util/userSettings/public.key
DIFSRIP: no, you should use the default keys. I guess it is not required to add the key here, can you check?

test/reuse/util/userSettings/setDateFormatFromUserSettings.spec.js
DIFSRIP: same as above as discussed yesterday

test/reuse/util/userSettings/setLanguageFromUserSettings.spec.js
```diff
@@ -0,0 +1,16 @@
...
+describe("userSettings", function () {
+  describe("setLanguageFromUserSettings.spec", function () {
+
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution & Verification", async function () {
+      await util.userSettings.setLanguageFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
+      common.assertion.expectEqual(process.env.USER_SETTINGS_LANG_KEY.length, 2);
```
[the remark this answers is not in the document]
DIFSRIP: I don't understand what is being tested here

test/reuse/util/userSettings/setS4UserSettings.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    describe("setS4UserSettings.spec - apply all s4 user settings via config<->login", function () {
+      it("Preparation: Set systemUrl and baseUrl ", async function () {
+        browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+        browser.config.baseUrl = util.data.decrypt(data.baseUrl);
+      });
+
+      it("Execution & Verification", async function () {
+        await ui5.navigation.navigateToApplication("BusinessUser-maintain");
+        await ui5.session.login(util.data.decrypt(data.username), util.data.decrypt(data.password));
+        common.assertion.expectEqual(DATE_FORMAT_LENGTH, process.env.USER_SETTINGS_DATE_FORMAT.length);
```
DIFSRIP, line 25: is it possible to add an actual assertion since this line is unrelated to the one before

DIFSRIP (changes requested):
Please see my comments

src/reuse/modules/util/userSettings.ts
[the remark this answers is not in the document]
Templada: information of which class is missing then

.gitignore
```diff
@@ -36,7 +36,7 @@ Dictionaries/
 GPUCache/
 exthost Crash Reports/
 Local Storage/
-User/
+[U]ser/
```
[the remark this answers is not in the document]
Templada: idk where this was from tbh, will remove it

.prettierrc.json
```diff
@@ -1,6 +1,6 @@
 {
     "tabWidth": 2,
-    "printWidth": 300,
+    "printWidth": 200,
```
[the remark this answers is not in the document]
Templada: reverted

test/reuse/util/userSettings/setDateFormatFromUserSettings.spec.js
Templada: as discussed js can not handle types as we expected in our meeting, therefore keeping the date evaluation only.

src/reuse/modules/util/formatter.ts
Templada: took it from main, as discussed we will refactor it with the suggested code in another PR

src/reuse/modules/util/formatter.ts
Templada: as said we will do that in another PR

test/reuse/util/userSettings/setS4UserSettings.spec.js
Templada: added expectDefined() as it is to generic to have test against static values.

Templada: regarding the Language Key comment: "I don't understand what is being tested here" -> The Language Key can only be 2 chars long therefore checking if it matches a length of 2

Templada:
@vlad9719: Can you please check and review again, everything should be resolved now.

Private key still don't work - agreed with Benny that we comment it out after successful run, since it is connecting to an internal system and is likely to fail due to connectivity issues (maint/downtimes/etc).

So if you want to check if the tests are fine/running green, please execute them locally.
All tests should run w/ success:
image
image

test/reuse/util/userSettings/setNumberFormatFromUserSettings.spec.js
```diff
@@ -0,0 +1,15 @@
...
+
+describe("userSettings", function () {
+  describe("setNumberFormatFromUserSettings.spec", function () {
+    
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution", async function () {
+      await util.userSettings.setNumberFormatFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
```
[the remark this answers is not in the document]
Templada: done

test/reuse/util/userSettings/setTimeFormatFromUserSettings.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe("userSettings", function () {
+  describe("setTimeFormatFromUserSettings.spec", function () {
+
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution", async function () {
+      await util.userSettings.setTimeFormatFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
```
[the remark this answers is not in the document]
Templada: done

test/reuse/util/userSettings/setTimeZoneFromUserSettings.spec.js
```diff
@@ -0,0 +1,15 @@
...
+
+describe("userSettings", function () {
+  describe("setTimeZoneFromUserSettings.spec", function () {
+
+    it("Preparation: Set systemUrl ", async function () {
+      browser.config.params.systemUrl = util.data.decrypt(data.systemUrl);
+    });
+
+    it("Execution", async function () {
+      await util.userSettings.setTimeZoneFromUserSettings(util.data.decrypt(data.username), util.data.decrypt(data.password));
```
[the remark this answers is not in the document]
Templada: done

test/reuse/util/userSettings/getTimeFormatFromUserSettings.spec.js
Templada: added assertion

test/reuse/util/userSettings/setLanguageFromUserSettings.spec.js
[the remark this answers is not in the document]
Templada: @vlad9719 : no since it could be changed on system, therefore only length of language key can be checked

test/reuse/util/userSettings/getDateFormatFromUserSettings.spec.js
Templada: as discussed in the meeting, type cant be checked for now as we dont use ts, therefore deleted the length check and kept the date verification

src/reuse/modules/util/userSettings.ts
Templada: move all api calls in _getXXXXresponse functions and added try catch. Error handling will be improved later on in different PR as we agreed

vlad9719 (approved):
Looks good!

PR #167, opened 2023-07-03
Parse txt and xml

Templada:


Templada:
Info: .txt will be handled separately in a different PR later on.

cvakiitho (dismissed):

cvakiitho:
I'm just not 100% sure if we shouldn't lock the test urls to newer version instead of unlocking it. 
But honestly probably better to unlock since then we'd get the possible problems first instead of fiori apps.

Templada:
> I'm just not 100% sure if we shouldn't lock the test urls to newer version instead of unlocking it. But honestly probably better to unlock since then we'd get the possible problems first instead of fiori apps.

I just checked and seems that adding the fiori theme to the old url fixes the issue.

But in general i agree with you. Rather we facing new issues before the user facing them.
Still I would like to open up for discussion and handle that in a different PR then.

PR #169, opened 2023-07-04
Add parse txt

Templada:


src/reuse/modules/util/file.ts
```diff
@@ -182,35 +182,59 @@ export class File {
...
    */
   async expectTextDataToContain(filePath: string, searchString: string): Promise<any> {
     const vl = this.vlf.initLog(this.expectTextDataToContain);
 
-    try {
-      const fileContent = fs.readFileSync(filePath, "utf-8");
-      common.assertion.expectTrue(fileContent.includes(searchString));
-    } catch (error) {
-      throw new Error("Function 'expectTextDataToContain' failed: Search String not included in .txt file.");
+    if (fs.existsSync(filePath) && this._checkFileEnding(filePath, "txt")) {
```
cvakiitho, line 216: are you sure this method is used only for .txt files?

cvakiitho (approved):

PR #176, opened 2023-07-18
Separate data encryption tests

heraldantony:
Separate tests using private key.
Adjust workflow so that these tests do not run on a pull request created from fork

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -152,7 +152,7 @@ jobs:
           npm run test:reuse:util
 
   test-util-data-privacy:
-    if: github.repository != github.event.pull_request.base.repo.full_name
+    if: github.repository == github.event.pull_request.base.repo.full_name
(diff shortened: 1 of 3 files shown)


.github/workflows/node.js.yml
```diff
@@ -135,6 +135,24 @@ jobs:
...
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: run tests
+        run: |
+          npm ci
+          npm run update-chromedriver
+          npm run test:reuse:util
+
+  test-util-data-privacy:
+    if: github.repository != github.event.pull_request.base.repo.full_name
```
DIFSRIP, line 155: I guess we have to reverse this logic sine they have to be executed for non forks. At this PR they are not
```suggestion
    if: github.repository == github.event.pull_request.base.repo.full_name
```

DIFSRIP (changes requested):

PR #191, opened 2023-08-02
Fixed Scroll Issue

ihappyk:
The scrollToView has the parameter which is of object by default the earlier implementation was taking both inline & block as  center.
as per the wdio by default its suppose to be inline as nearest and block to be start.

this pr contains the above changes


DIFSRIP (approved):
looks good

PR #192, opened 2023-08-03
Replaced addValue to setValue for file upload

Ananthreddy3004:
const remoteFilePath = await browser.uploadFile(filePath);
await elem.addValue(remoteFilePath);  

addValue method will not clear the textbox so in first iteration it will upload one file **test.txt** and in second iteration it will upload new file **test1.txt** along with **test.txt** because this file already exist in textbox so we can see duplicate of file upload.

if we use the below this things in place of addValue it will work as expected
await elem.setValue(remoteFilePath);

### DIFF
--- a/src/reuse/modules/util/file.ts
+++ b/src/reuse/modules/util/file.ts
@@ -44,7 +44,7 @@ export class File {
         const filePath = path.resolve(file);
         vl.log(`Uploading file with a path ${filePath}`);
         const remoteFilePath = await browser.uploadFile(filePath);
-        await elem.addValue(remoteFilePath);
+        await elem.setValue(remoteFilePath);
       }
     } catch (error) {
       throw new Error(`Function 'upload' failed': ${error}`);


heraldantony (approved):
looks good

src/reuse/modules/util/file.ts
```diff
@@ -44,7 +44,7 @@ export class File {
         const filePath = path.resolve(file);
         vl.log(`Uploading file with a path ${filePath}`);
         const remoteFilePath = await browser.uploadFile(filePath);
-        await elem.addValue(remoteFilePath);
+        await elem.setValue(remoteFilePath);
```
DIFSRIP, line 47: @heraldantony mentioned, that setValue is only possible for displayed elements. Are we sure it is working with that change for hidden inputs?

Ananthreddy3004: @DIFSRIP for hidden element explicitly they have handle it externally to make visible samething we followed in customerE2E.if we want me to handle it at util level then I will do it.

DIFSRIP: are there many such cases where it is not visible, do we have concrete samples?

DIFSRIP: We can improve it in a separate PR if so, need to release the fix soon, so I will merge it like this for now.

PR #196, opened 2023-08-09
Fix util.file.upload for cases where file input is not visible or interactable

heraldantony:
elem.setValue gives "element not interactable" error if file input is not visible in chrome.
In firefox changing visibility doesn't help, so switched to elem.addValue


Ananthreddy3004 (approved):

DIFSRIP (approved):
looks good

PR #200, opened 2023-08-14
Update types.ts

ihappyk:


DIFSRIP (commented):
Where did you got that from, that behaviour is a valid option parameter?

Official docs says nothing about it: https://webdriver.io/docs/api/element/scrollIntoView/

Also if we introduce that, we have to adjust the functions using it as well.

Please clarify.

DIFSRIP:
@ihappyk closing this PR, feel free to reopen if required

PR #201, opened 2023-08-14
Added customised parameter as option for logException method

Ananthreddy3004:
Handled the additional parameter as optional parameter for displaying the custom error message and stacktrace.

Syntax:
errorHandler.logException(errorObject: unknown | Error, customErrorMessage?: string):

Example:
errorHandler.logException(new Error(), "Please pass the correct locator"):
errorHandler.logException(new Error("message")):

### DIFF
--- a/src/reuse/helper/errorHandler.ts
+++ b/src/reuse/helper/errorHandler.ts
@@ -2,7 +2,7 @@ import chalk from "chalk";
 import { ErrorMessages } from "../helper/errorMessages";
 
 export interface IErrorHandler {
-  logException(error: Error): Promise<never>;
+  logException(error: unknown | Error, customErrorMessage?: string): Promise<never>;
 }
 
 export class CustomError extends Error {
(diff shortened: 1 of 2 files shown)


src/reuse/helper/errorHandler.ts
```diff
@@ -53,8 +55,8 @@ export default class ErrorHandler implements IErrorHandler {
 
   // =================================== HELPER ===================================
 
-  private _retrieveFunctionNameFromStack(errorObject: Error): string {
-    if (errorObject.stack) {
+  private _retrieveFunctionNameFromStack(errorObject: unknown | Error): string {
+    if (errorObject instanceof Error && errorObject.stack) {
```
DIFSRIP, line 59: as discussed in our call, function name should be extracted from lowest level of the call stack meaning the initial function the use has called. But you can do it in a separate PR.

DIFSRIP (approved):

PR #208, opened 2023-08-23
fix generate-types command

Varun-Hegde:
Below issues have been fixed:
- ```npm run generate-types``` was not working on windows due to the path separator
- individual type declarations files which are not inside any modules are not being handled properly leading to generation of types.t.d folder


DIFSRIP (commented):
Looks good so far! Will test it with the next release before merging it.

DIFSRIP (approved):

DIFSRIP (approved):

PR #229, opened 2023-12-20
added update-docs workflow

vlad9719:


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,19 @@
+name: Node.js CI
```
cvakiitho, line 1: I'm not sure if two workflows can have same name, could you please change it to Docs Deployment, or something?

vlad9719: done

cvakiitho (approved):

cvakiitho:
LGTM thanks!

PR #232, opened 2023-12-22
update pkgs for node v20

heraldantony:
Update packages to fix "SyntaxError: Cannot use import statement outside a module" when using TypeScript.


DIFSRIP (approved):
Looks good

PR #234, opened 2024-01-12
Support optional headers for OData calls

DIFSRIP:
- Adds support for optional headers passed to the OData _init_ and _HTTP_ methods
- Cleaned up the OData service file

### DIFF
--- a/src/reuse/modules/service/odata.ts
+++ b/src/reuse/modules/service/odata.ts
@@ -89,9 +89,9 @@ export class OData {
     const auth: any = {
       type: authType ?? "",
       username: username,
-      password: password,
-      headers: headers
+      password: password
     };
+    if (headers && Object.entries(headers).length > 0) auth.headers = headers;
 
     const srv = new this.Service({
       logger: loggingEnabled ? logger : "",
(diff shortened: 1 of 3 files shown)


src/reuse/modules/service/odata.ts
```diff
@@ -55,31 +78,29 @@ export class OData {
...
 
-    if (authType) {
-      auth["type"] = authType;
-    }
     const srv = new this.Service({
       logger: loggingEnabled ? logger : "",
       url: url,
       auth: auth,
-      parameters: parameters, //Define initial request by $metadata?sap-client=<client-number>&sap-documentation=&sap-language=EN
+      parameters: parameters, // Define initial request by $metadata?sap-client=<client-number>&sap-documentation=&sap-language=EN
```
vlad9719, line 100: Just a suggestion here, not critical: use the shorthand property assignment, when value for some key inside an object is taken from a variable with the same name as the key has

`const src = ...{
  ...,
  url,
  auth,
  parameters
}`

More info just in case: https://ui.dev/shorthand-properties

vlad9719 (dismissed):
Looks good to me. I've added a small optional suggestion, please see the comment

src/reuse/modules/service/odata.ts
DIFSRIP: Sure thing! Will change it, thanks!

DIFSRIP: done

vlad9719 (approved):

PR #236, opened 2024-01-16
Login via env var

DIFSRIP:
- Add support to use env vars for username and password.
- Requested to centrally run tests via STEP using custom users: https://github.tools.sap/sProcurement/qmate/issues/476

TODO:

- [x] Test
- [x] Update Docs


Varun-Hegde (dismissed):
Looks Good!

PR #247, opened 2024-01-25
update odata lib dependency

DIFSRIP:


Templada:
related to [https://github.tools.sap/gs-fin/qmate-consultation-hour/issues/59](APS_IAM_CUSR_SRV could not be initialized: Cannot read properties of undefined (reading 'reduce')) and [fix v2 initi](https://github.com/SAP/odata-library/pull/120)

PR #251, opened 2024-02-05
fix reuse retry handling

DIFSRIP:
- Fixes setting reuse retries via config params.
- Fixes logging retry loop index.


Varun-Hegde (approved):
looks good

PR #277, opened 2024-07-05
removed dependencies from browser object in before hooks where possible

vlad9719:
**What has been done**

Adjusted the `beforeSession` hook to check for the presence of a `browser` object

**Why**

To facilitate possibility of running test cases in the _watch_ mode. Without that check, the running of a test would fail because the `browser` global object would be undefined in the `beforeSession` hook, causing the code inside the hook to fail, when a test case is restarted in the _watch_ mode


heraldantony (approved):
looks good

src/scripts/dataExchange/dataExchange.ts
```diff
@@ -7,64 +7,37 @@
...
-        // @ts-ignore
-        await importExportDataUtil.readData(fileOrDir, [param]);
-      } else {
-        delete importParams[param];
-        console.warn(
-          `"${fileOrDir}" does not exist or is not readable. Please check path or permissions.`
-        );
-      }
+    const importParams = config.params.import || {};
+    if (importParams) {
```
DIFSRIP, line 29: this can never be falsy because of {}

src/scripts/dataExchange/dataExchange.ts
```diff
@@ -212,3 +185,51 @@ class DataExchange {
...
+        delete importParams[param];
+        console.warn(
+          `"${fileOrDir}" does not exist or is not readable. Please check path or permissions.`
+        );
+      }
+    } 
+}
+
+async function readExportParams(config: Record<string, any>, exportParams: any) {
+  if (!exportParams) {
```
DIFSRIP, line 212: check is redundant: https://github.com/SAP/wdio-qmate-service/pull/277/files?diff=split&w=0#diff-bb5bafb01127841195d1c3c05a6ff59f0eb078f7ec1cb36892c2f179b6a5491dR38

DIFSRIP (changes requested):
Please see my comments

src/scripts/dataExchange/dataExchange.ts
vlad9719: removed the check

DIFSRIP (approved):

PR #286, opened 2024-09-10
Bump odata lib version

Templada:


Templada:
@cvakiitho: I'll do some testing before we merge

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sap_oss/wdio-qmate-service",
-  "version": "2.9.8",
+  "version": "2.9.9",
   "lockfileVersion": 3,
```
Templada, line 4: @cvakiitho gh-action complains about the lockversion. But actually it didn't get changed?

PR #287, opened 2024-09-11
Initial reusable functions for Mobile Native Automation

sirikumarsap:


package.json
```diff
@@ -109,7 +109,9 @@
     "test:reuse:service": "npm run test:reuse:service:rest && npm run test:reuse:service:odata",
     "test:reuse": "npm run test:reuse:common && npm run test:reuse:nonUi5 && npm run test:reuse:ui5 && npm run test:reuse:util && npm run test:reuse:service:odata",
     "test:reuse:helper:elementHighlight": "npx wdio ./test/reuse/helper/elementHighlight/test.elementHighlight.conf.js",
-    "test": "npm run test:authenticator:staticLogin && npm run test:core && npm run test:reuse"
+    "test": "npm run test:authenticator:staticLogin && npm run test:core && npm run test:reuse",
+    "device:local": "npx wdio wdio.local.android.device.conf.js",
```
Templada, line 113:
```suggestion
    "android:local": "npx wdio wdio.local.android.device.conf.js",
```

package.json
```diff
@@ -109,7 +109,9 @@
...
+    "device:config": "npx wdio ./test/reuse/mobile/element/test.locator.local.android.device.conf.js"
```
Templada, line 114:
```suggestion
    "android:config": "npx wdio ./test/reuse/mobile/element/test.locator.local.android.device.conf.js"
```

package.json
```diff
@@ -128,13 +130,17 @@
...
+    "appium-uiautomator2-driver": "^3.7.4",
+    "appium-xcuitest-driver": "^7.24.3",
     "axios": "^1.6.0",
     "deepmerge": "^4.2.2",
     "fs-extra": "^10.0.0",
     "path": "^0.12.7",
     "pdf-parse": "^1.1.1",
     "query-string": "^7.1.0",
     "supertest": "^7.0.0",
+    "wdio-chromedriver-service": "^8.0.0",
```
Templada, line 143: I am not sure if that should be added back. @cvakiitho I think we removed the chromedriver package for a reason, do still remember?

src/reuse/modules/mobile/Mobile.ts
```diff
@@ -0,0 +1,14 @@
+import gestures, { Gestures } from "./gestures";
+import element, { ElementModule } from "./element";
+import userInteraction, { UserInteraction } from "./userInteraction";
+import device, { Device } from "./device";
+
+export class Mobile {
+  gestures: Gestures = gestures;
+  userInteraction: UserInteraction = userInteraction;
+  element: ElementModule = element;
```
Templada, line 9: is 'Element' protected?

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,119 @@
...
+  private vlf = new VerboseLoggerFactory("mobile", "device");
+  private ErrorHandler = new ErrorHandler();
+
+  /**
+   * @function isAppInstalled
+   * @memberOf mobile.device
+   * @description Check wether given package/bundle app is installed or not in the device.
+   * @param {string} packageIdorBundleId - Android package Id, or iOS bundle Id.
+   * @returns {boolean} Returns true if specified app package/bundled installed in the device, or false.
+   * @example await mobile.device.isAppInstalled("com.google.android.apps.maps");
```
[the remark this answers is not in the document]
sirikumarsap: No, its an example, in general every Android/iOS app's can be recognised with its package name(Android) or bundle id(iOS).

src/reuse/modules/util/browser.ts
```diff
@@ -417,6 +417,46 @@ export class Browser {
...
+  }
+
+  /**
+   * @function isIOS
+   * @memberOf util.browser
+   * @description Indicates an iOS session
+   * @returns {boolean} Return true if its a iOS session driver.
+   * @example await util.browser.isIOS();
+   */
+  async isIOS(): Promise<boolean> {
```
Templada, line 454: Would go with strict camelCase here - @cvakiitho @vlad9719 wdy think?

```suggestion
  async isIos(): Promise<boolean> {
```

test/reuse/mobile/element/test.locator.local.android.device.conf.js
```diff
@@ -0,0 +1,349 @@
+const path = require("path");
+const merge = require("deepmerge");
+const profile = require("../../../helper/configurations/mobile.conf");
+
+exports.config = merge(profile.config, {
+    //
```
Templada, line 6: actually would only keep the necessary configuration options and delete the comments

Templada (changes requested):
everything looks good so far, besides a few uncertainties.

Please add more tests for every single function which is testable in the new modules

Templada:
@vlad9719 please take a look as well

src/reuse/modules/util/browser.ts
vlad9719: both are fine as for me

package.json
```diff
@@ -109,7 +109,9 @@
     "test:reuse:service": "npm run test:reuse:service:rest && npm run test:reuse:service:odata",
     "test:reuse": "npm run test:reuse:common && npm run test:reuse:nonUi5 && npm run test:reuse:ui5 && npm run test:reuse:util && npm run test:reuse:service:odata",
     "test:reuse:helper:elementHighlight": "npx wdio ./test/reuse/helper/elementHighlight/test.elementHighlight.conf.js",
-    "test": "npm run test:authenticator:staticLogin && npm run test:core && npm run test:reuse"
+    "test": "npm run test:authenticator:staticLogin && npm run test:core && npm run test:reuse",
+    "android:local": "npx wdio wdio.local.android.device.conf.js",
```
vlad9719, line 113: Probably it's better to follow convention used in the other npm commands here: `test:mobile:android:local`

package.json
```diff
@@ -109,7 +109,9 @@
...
+    "android:config": "npx wdio ./test/reuse/mobile/element/test.locator.local.android.device.conf.js"
```
vlad9719, line 114: Same here: `test:mobile:android:config`

package.json
vlad9719: Also, the command doesn't run, I guess it's because config file route is incorrect. You would either need to add a file (and then place it to the same folder as the other config), or remove the command

src/reuse/modules/mobile/Mobile.ts
```diff
@@ -0,0 +1,14 @@
+import gestures, { Gestures } from "./gestures";
+import element, { ElementModule } from "./element";
+import userInteraction, { UserInteraction } from "./userInteraction";
+import device, { Device } from "./device";
+
+export class Mobile {
+  gestures: Gestures = gestures;
+  userInteraction: UserInteraction = userInteraction;
```
vlad9719, line 8: Is there a significant difference between actions in the `gestures`and `userInteraction`namespaces? I am not a mobile tester, just curious if it's really better to have these two different namespaces instead of one

test/reuse/mobile/element/isPresent.spec.js
```diff
@@ -0,0 +1,20 @@
+//var assert = require("assert");
```
vlad9719, line 1: As far as I understood, there are no assertions in this test case, it just taps a button. Could you add some assertions here? An example can be found here:
https://github.com/SAP/wdio-qmate-service/blob/main/test/reuse/ui5/userInteraction/click.spec.js
Or you can also explore the other test cases in the test folder and find examples which suit here better

test/reuse/mobile/element/test.locator.local.android.device.conf.js
```diff
@@ -0,0 +1,349 @@
...
+     * Function to be executed after a test (in Mocha/Jasmine only)
+     * @param {object}  test             test object
+     * @param {object}  context          scope object the test was executed with
+     * @param {Error}   result.error     error object in case the test fails, otherwise `undefined`
+     * @param {*}       result.result    return object of test function
+     * @param {number}  result.duration  duration of test
+     * @param {boolean} result.passed    true if test has passed, otherwise false
+     * @param {object}  result.retries   information about spec related retries, e.g. `{ attempts: 0, limit: 0 }`
+     */
+    afterTest: async function (test, context, { error, result, duration, passed, retries }) {
```
vlad9719, line 277: Is this hook necessary only for this config, or it can be required also in other tests in the repo, and therefore should be moved to the parent config `mobile.conf`?

vlad9719 (changes requested):
Looks good but a little bit incomplete, since, as Marvin said, tests are needed for every added function

Also please see some other suggestions from my side

package.json
sirikumarsap: The appium-uiautomator2-driver requires the wdio-chromedriver-service because it helps facilitate seamless communication between Appium and the Chrome browser when running automated tests. This service starts the Appium server in a child process, allowing you to manage ChromeDriver instances effectively, which is essential for testing web applications on Android devices via Appium with WebdriverIO (WDIO).

src/reuse/modules/mobile/Mobile.ts
sirikumarsap: Mobile gestures and user interactions are closely related but not the same. Mobile gestures refer specifically to touch-based commands, like tapping, swiping, or pinching, that users perform on a device to execute actions. User interactions encompass a broader range of ways users engage with the app, including gestures as well as other forms of input like button clicks or voice commands. 

Essentially, gestures are a subset of user interactions.

Refer to below article for more info:
https://www.justinmind.com/blog/app-interaction-gestures/#:~:text=interactions%20in%20Justinmind-,Difference%20between%20app%20interactions%20and%20gestures,users%20gesticulate%20and%20apps%20interact.

PR #294, opened 2024-09-17
Initial reusable functions for Mobile Automation

sirikumarsap:


package.json
```diff
@@ -128,13 +129,17 @@
...
+    "appium-uiautomator2-driver": "^3.7.4",
+    "appium-xcuitest-driver": "^7.24.3",
     "axios": "^1.6.0",
     "deepmerge": "^4.2.2",
     "fs-extra": "^10.0.0",
     "path": "^0.12.7",
     "pdf-parse": "^1.1.1",
     "query-string": "^7.1.0",
     "supertest": "^7.0.0",
+    "wdio-chromedriver-service": "^8.0.0",
```
sirikumarsap, line 142: @Templada : Previous comment in #287 
I am not sure if that should be added back. @cvakiitho I think we removed the chromedriver package for a reason, do still remember?

sirikumarsap: The appium-uiautomator2-driver requires the wdio-chromedriver-service because it helps facilitate seamless communication between Appium and the Chrome browser when running automated tests. This service starts the Appium server in a child process, allowing you to manage ChromeDriver instances effectively, which is essential for testing web applications on Android devices via Appium with WebdriverIO (WDIO).

src/reuse/modules/mobile/Mobile.ts
```diff
@@ -0,0 +1,13 @@
+import gestures, { Gestures } from "./gestures";
+import element, { ElementModule } from "./element";
+import userInteraction, { UserInteraction } from "./userInteraction";
+import device, { Device } from "./device";
+
+export class Mobile {
+  gestures: Gestures = gestures;
+  userInteraction: UserInteraction = userInteraction;
+  element: ElementModule = element;
```
sirikumarsap, line 9: @Templada Previous comment in #287 
is 'Element' protected?

sirikumarsap: it's just like same web, /Users/i542683/GitHub/qmate_bma_17092024/wdio-qmate-service/src/reuse/modules/nonUi5/element.ts

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,116 @@
...
+  private vlf = new VerboseLoggerFactory("mobile", "device");
+  private ErrorHandler = new ErrorHandler();
+
+  /**
+   * @function isAppInstalled
+   * @memberOf mobile.device
+   * @description Check wether given package/bundle app is installed or not in the device.
+   * @param {string} packageIdorBundleId - Android package Id, or iOS bundle Id.
+   * @returns {boolean} Returns true if specified app package/bundled installed in the device, or false.
+   * @example await mobile.device.isAppInstalled("com.google.android.apps.maps");
```
sirikumarsap, line 20: @Templada : Previous comment in #287 
so this is checking for google maps to be installed?

sirikumarsap: No, its an example, in general every Android/iOS app's can be recognised with its package name(Android) or bundle id(iOS).

package.json
sirikumarsap: if it creates any issues at this service(wdio-qmate-service) level, we remove this, and add at our consume this service level.

src/reuse/modules/util/browser.ts
```diff
@@ -417,6 +417,46 @@ export class Browser {
...
+  }
+
+  /**
+   * @function isIOS
+   * @memberOf util.browser
+   * @description Indicates an iOS session
+   * @returns {boolean} Return true if its a iOS session driver.
+   * @example await util.browser.isIOS();
+   */
+  async isIOS(): Promise<boolean> {
```
sirikumarsap, line 454: @Templada : Previous comment #287 

Would go with strict camelCase here - @cvakiitho @vlad9719 wdy think?

src/reuse/modules/util/browser.ts
```diff
@@ -417,6 +417,45 @@ export class Browser {
...
+  }
+
+  /**
+   * @function isiOS
+   * @memberOf util.browser
+   * @description Indicates an iOS session
+   * @returns {boolean} Return true if its a iOS session driver.
+   * @example await util.browser.isiOS();
+   */
+  async isiOS(): Promise<boolean> {
```
sirikumarsap, line 454: @Templada Previous comment #287 
Would go with strict camelCase here - @cvakiitho @vlad9719 wdy think?

sirikumarsap: As per the Apple standard of using this name as iOS(i-small letter, OS -would be capital) now it's changed to as per this standard.

vlad9719: The version from Siri is fine for me

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,116 @@
...
+   * @param {string} appPath - Path of the app(.apk, .ipa)
+   * @example 
+   * await mobile.device.installAppForPlatform("/path/to/your/app.apk");
+   * await mobile.device.installAppForPlatform("/path/to/your/app.ipa");
+   */
+async installAppForPlatform(appPath: string): Promise<void> {
+  const vl = this.vlf.initLog(this.installAppForPlatform);
+  const platform = await browser.capabilities.platformName;
+  try {
+      if (platform.toLowerCase() === 'android') {
```
vlad9719, line 46: The code in each block of the if-else statement is exactly the same, do you really need that if-else statement to check for the platform? And also, since the installation logic is the same for each platform, you might also need to name the method as just `installApp`

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,116 @@
...
+   * await mobile.device.pressKeyByNameInAndroid("enter");
+   * await mobile.device.pressKeyByNameInAndroid("space");
+   * await mobile.device.pressKeyByNameInAndroid("delete");
+   * await mobile.device.pressKeyByNameInAndroid("menu");
+   * await mobile.device.pressKeyByNameInAndroid("search");
+   * await mobile.device.pressKeyByNameInAndroid("camera");
+   * await mobile.device.pressKeyByNameInAndroid("focus");
+   * await mobile.device.pressKeyByNameInAndroid("notification");
+   */
+async pressKeyByNameInAndroid(keyName: string): Promise<void> {
```
vlad9719, line 83: One thought to consider: maybe it would be better to organize the whole mobile namespace in a way that we have a sub namespace for each mobile platform? Then it would be `mobile.android.pressKeyByName` and `mobile.ios.pressKeyByName`. This would make sense to me, but I don't know all the nuances of mobile testing. @sirikumarsap @Templada what do you think?

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    const keyCode = keyCodeMap[keyName.toLowerCase()];
+    if (keyCode !== undefined) {
+      vl.log(`Pressing key: ${keyName} (code: ${keyCode})`);
+      await browser.pressKeyCode(keyCode);
+    } else {
+        throw new Error(`Key code for "${keyName}" not found.`);
+    }
+  }catch(error){
+    return this.ErrorHandler.logException(error, 'Error: During pressKeyByNameInAndroid in Android', true);
```
vlad9719, line 111:
```suggestion
    return this.ErrorHandler.logException(error, 'Error: During pressKeyByNameInAndroid', true);
```
Since there is already word 'Android' in the function name

vlad9719 (changes requested):
Thank you for your contributions!

Please consider the suggestions from my side

One more thing: please add at least one test for each of the functions you have added

src/reuse/modules/mobile/device.ts
Templada: yeah, i meant that example if that was the package name of maps?

src/reuse/modules/mobile/device.ts
Templada: agree, would put every check in one single if statement if any of the platforms are matched -> if so, install the app on the supported plattform and print a message that it was successfully installed on the given platform.

src/reuse/modules/mobile/device.ts
Templada: agree, everything which is specific for a certain platform should be moved to the corresponding (new) namespace -> mobile.ios.; mobile.android and so on. For functions which are platform independent or same we should have a module named common

@sirikumarsap

Templada: interesting to know would be if the elements also need to be located differently from platform to platform. 

If so we would need another submodules as well under mobile.<platform>.element

src/reuse/modules/util/browser.ts
Templada: as I remember I agreed with Ben that we wanted to have true camel-case.
Per reference we also went same approach with `json` in [function names](https://github.com/SAP/wdio-qmate-service/blob/a189ae33828c0c7351b7246b1c06f20d9dcd08ca/src/scripts/dataExchange/dataExchangeUtil.ts#L62)

Templada (changes requested):
please see the updated comments

src/reuse/modules/mobile/device.ts
DIFSRIP: you can dynamically set the log using the platform variable and avoid the differentiation as the above comments already explained.

src/reuse/modules/mobile/device.ts
DIFSRIP: same here, maybe good to separate by android/ios namespace

src/reuse/modules/mobile/gestures.ts
```diff
@@ -0,0 +1,38 @@
...
+  /**
+ * Swipe from one point to another on the screen.
+ * @param {number} startX - The starting X coordinate of the swipe
+ * @param {number} startY - The starting Y coordinate of the swipe
+ * @param {number} endX - The ending X coordinate of the swipe
+ * @param {number} endY - The ending Y coordinate of the swipe
+ * @param {number} duration - The duration of the swipe in milliseconds (optional, default is 1000ms)
+ * @returns {Promise<void>}
+ */
+async swipe(startX: number, startY: number, endX: number, endY: number, duration: number = 1000): Promise<void> {
```
DIFSRIP, line 21: not sure if it needs a separate namespace for gestures since it is also a kind of userinteraction.

src/reuse/modules/util/browser.ts
DIFSRIP: Yes I know that is a complicated topic and there are different opinions on that. In general, there is no right ore wrong.

However, as Marvin said, using strict camel case even for terms like "OS" is the only way to separate the words correctly in a consistent way. 
Imagine you have something like "get OS info", if you would keep the OS, it is not clear where the separation happens -> "getOSInfo"

package.json
DIFSRIP: ok, I guess we need these dependecies for testing only, if so, please move them to dev dependecies.

DIFSRIP (changes requested):
Thanks a lot for your contribution. I have added some comments.
We can discuss on them in our meeting. Let me know if you need any support to speed things up.

test/helper/configurations/mobile.conf.js
```diff
@@ -0,0 +1,15 @@
...
+
+exports.config = merge(baseConfig.config, {
+
+  port: 4723,
+
+  services: [
+    ['appium', {
+      logPath: './logs',
+      logFileName: 'qmate-appium.log',
+    }]
```
DIFSRIP, line 12: add chromedriver service as in profiles

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,116 @@
...
+      delete: 67,
+      menu: 82,
+      search: 84,
+      camera: 27,
+      focus: 80,
+      notification: 83,
+  };
+
+  try {
+    const keyCode = keyCodeMap[keyName.toLowerCase()];
```
DIFSRIP, line 103: add if-else for arg type (string, number) to support both, name and number key

DIFSRIP (changes requested):

src/reuse/modules/mobile/device.ts
```diff
@@ -0,0 +1,116 @@
...
+async installAppForPlatform(appPath: string): Promise<void> {
+  const vl = this.vlf.initLog(this.installAppForPlatform);
+  const platform = await browser.capabilities.platformName;
+  try {
+      if (platform.toLowerCase() === 'android') {
+          vl.log("Installing Android app...");
+          await browser.installApp(appPath);
+          vl.log("Android app installed successfully.");
+      } else if (platform.toLowerCase() === 'ios') {
+          vl.log("Installing iOS app...");
```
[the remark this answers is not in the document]
sirikumarsap: Done, Added logic with single if condition.

src/reuse/modules/mobile/device.ts
sirikumarsap: Done, Added logic with single if condition.

src/reuse/modules/mobile/device.ts
sirikumarsap: As per this review comments, these specific functions are now moved to device platform-wise, now mobile.android
mobile.ios 
files are created.

src/reuse/modules/mobile/device.ts
sirikumarsap: Done

src/reuse/index.ts
```diff
@@ -108,6 +109,21 @@ class ReuseLibrary {
...
     };
+
+     /**
+     * @global
+     * @description Global namespace for mobile modules.
+     */
+     const mobile = {
+      element: mobileQmate.element,
+      userInteraction: mobileQmate.userInteraction,
+      gestures: mobileQmate.gestures,
```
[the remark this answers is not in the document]
sirikumarsap: As we discussed we are keeping the gestures class

src/reuse/modules/mobile/device.ts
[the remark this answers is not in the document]
sirikumarsap: Now, it added two specific classes for these mobile platforms, and this part of the code is moved to https://github.com/SAP/wdio-qmate-service/blob/2cfce035e869229454bee0785ac403e49b97ab07/src/reuse/modules/mobile/android.ts

src/reuse/modules/mobile/device.ts
sirikumarsap: Added a new function that accepts a number as a parameter as a key code.
https://github.com/SAP/wdio-qmate-service/blob/2cfce035e869229454bee0785ac403e49b97ab07/src/reuse/modules/mobile/android.ts

src/reuse/modules/util/browser.ts
sirikumarsap: Changed as per detailed review comments.

src/reuse/modules/util/browser.ts
sirikumarsap: Yes, now its changed with strict camel case :)

src/reuse/modules/mobile/android.ts
```diff
@@ -0,0 +1,91 @@
+import { Element } from "../../../../@types/wdio";
+import { VerboseLoggerFactory } from "../../helper/verboseLogger";
+import ErrorHandler from "../../helper/errorHandler";
+/**
+ * @class Android
+ * @memberof Mobile
```
DIFSRIP, line 6: I guess the @memberof value has to be lowercase since the final namespaces are always lowercase. Otherwise the docs will not be mapped correctly.
```suggestion
 * @memberof mobile
```

DIFSRIP: same everywhere

DIFSRIP: we can do the changes in our call and verify the docs together

PR #295, opened 2024-09-18
fix wrong typ assignment in_decryptRecursively()

Templada:


Templada:
I was not part of the development of the decrypt function and therefore i don't know why it was designed like this.
Actually it seems that the if statement was like this since the beginning. But now our actions started to fail, and we need to fix it.

Although it looks wrong to me.

Could you please take a look as well @cvakiitho, @vlad9719 ?

src/reuse/modules/util/data.ts
```diff
@@ -136,7 +136,7 @@ export class Data {
     for (const key in data) {
       if (typeof data[key] === "object" && !Array.isArray(data[key])) {
         data[key] = this._decryptRecursively(data[key]);
-      } else if (typeof data[key] === "string" || Array.isArray(data[key])) {
+      } else if (typeof data[key] === "string") {
```
vlad9719, line 139: Basically, the check is correct, since the `global.util.data.decrypt` accepts either a string or an array of strings. Check the function [here](https://github.com/SAP/wdio-qmate-service/blob/a189ae33828c0c7351b7246b1c06f20d9dcd08ca/src/scripts/hooks/utils/decryption.ts#L45)

It would be more correct to add a right type to the global decrypt function [here](https://github.com/SAP/wdio-qmate-service/blob/a189ae33828c0c7351b7246b1c06f20d9dcd08ca/src/reuse/modules/util/Util.ts#L11)

I would suggest that you make it as
```ts
  decrypt: (input: string | Array<string>) => string;
```

Then type check and tests should work fine

vlad9719 (changes requested):
Please see my comment

src/reuse/modules/util/data.ts
vlad9719: I think we might also need to update the docs accordingly [here](https://github.com/SAP/wdio-qmate-service/blob/a189ae33828c0c7351b7246b1c06f20d9dcd08ca/src/reuse/modules/util/data.ts#L124)

DIFSRIP:
> I am currently checking on that, not sure if the change will work. Please do not merge yet.

@vlad9719 already mentioned that correct fix. The initial type check is essential, seems like only the interface was not adjusted and that never came up somehow.

DIFSRIP (approved):

DIFSRIP (approved):

PR #318, opened 2024-12-13
docs: open external links in a new tab

vlad9719:
This PR adjusts the documentation so that the links which lead to external resources are getting open in a new tab. This is intended to enhance usability of the documentation


docs/doc.md
DIFSRIP, line 1: why is this file updated?

vlad9719: I guess it was generated as a result of the `npm run serve-docs` command

DIFSRIP (approved):

PR #362, opened 2025-04-15
change tests to use latest ui5 samples and demo app

heraldantony:
Since newer ui5 samples use iframe, add switchToIframe where required.


heraldantony:
The clearBrowser test doesn't seem to clear the cookies, the test doesn't capture this case. It is a known issue, have been using reloadSession to fully clear the browser.
Other than that this PR just updates tests, shouldn't affect qmate-service.

Templada (approved):

PR #363, opened 2025-04-28
fix: fix parameter arg naming in init odata

DIFSRIP:
- There was an issue in the odata init function because of a wrong naming in the `parameters` argument


Templada (dismissed):

DIFSRIP:
@Templada need some time to test tomorrow after a small change

Templada (approved):

PR #365, opened 2025-04-29
update logout function

vlad9719:
The `ui5.session.logout` method has been adjusted to support the new nonUi5 sign out button


src/reuse/modules/ui5/session.ts
```diff
@@ -336,16 +336,28 @@ export class Session {
...
     const vl = this.vlf.initLog(this._clickSignOut);
-    const selector = {
-      elementProperties: {
-        metadata: "sap.m.StandardListItem",
-        mProperties: {
-          id: "*logoutBtn"
+    
+    await Promise.any([scrollAndClickLogoutUi5(), scrollAndClickLogoutNonUi5()]);
+ 
+    async function scrollAndClickLogoutUi5() {
```
DIFSRIP, line 342: can you change the names or add a comment to make it clear that this is simultaneously trying to call both (old and new) logout dialogs

I think, the ui5 and nonUi5 prefixes might confuse others and they could think it tries calling the same element using different technologies

DIFSRIP (changes requested):
Please see my comment

src/reuse/modules/ui5/session.ts
vlad9719: done

DIFSRIP (approved):

PR #376, opened 2025-05-12
feat: add a function to assert text value

uaroragit:
- add `expectTextToBe` to nonUi5 namespace

- add tests for 'expectTextToBe'

### DIFF
--- a/src/reuse/modules/nonUi5/assertion.ts
+++ b/src/reuse/modules/nonUi5/assertion.ts
@@ -82,6 +82,21 @@ export class Assertion {
     return common.assertion.expectEqual(value, compareValue);
   }
 
+  /**
+   * @function expectTextToBe
+   * @memberOf nonUi5.assertion
+   * @description Expects the text of the passed element to be the compare value.
+   * @param {Element | string} elementOrSelector - The element or CSS selector describing the element.
+   * @param {String} compareValue - The compare value.
+   * @example const element = await nonUi5.element.getById("button01");
+   **/
+  async expectTextToBe(elementOrSelector: Element | string, compareValue: string): Promise<void> {
+    const vl = this.vlf.initLog(this.expectTextToBe);
+    const element = await resolveCssSelectorOrElement(elementOrSelector);
+    const textValue = await element.getText();
+    return common.assertion.expectEqual(textValue, compareValue);
+  }
+
   // =================================== VISIBILITY ===================================
   /**
    * @function expectToBeVisible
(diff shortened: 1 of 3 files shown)


src/reuse/modules/nonUi5/assertion.ts
```diff
@@ -82,6 +82,21 @@ export class Assertion {
     return common.assertion.expectEqual(value, compareValue);
   }
 
+  /**
+   * @function expectTextToBe
+   * @memberOf nonUi5.assertion
+   * @description Expects the text of the passed element to be the compare value.
+   * @param {Element | string} elementOrSelector - The element or CSS selector describing the element.
+   * @param {String} compareValue - The compare value.
+   * @example const element = await nonUi5.element.getById("button01");
```
vlad9719, line 91: Please make the example full and include the subsequent use of the new method

src/reuse/modules/nonUi5/assertion.ts
```diff
@@ -82,6 +82,21 @@ export class Assertion {
...
+   **/
+  async expectTextToBe(elementOrSelector: Element | string, compareValue: string): Promise<void> {
+    const vl = this.vlf.initLog(this.expectTextToBe);
+    const element = await resolveCssSelectorOrElement(elementOrSelector);
+    const textValue = await element.getText();
```
vlad9719, line 96: As per the [WDIO doc](https://webdriver.io/docs/api/element/getText/) the function would return an empty string if the element is disabled or not visible. In that case, they recommend using [getHTML](https://webdriver.io/docs/api/element/getHTML/) method as a workaround

So I would propose the following change. @DIFSRIP let me know what you think

```suggestion
    const [elementText, elementHtml] = await Promise.all([element.getText(), element.getHTML()]);
    const textValue = elementText || elementHtml;
```

vlad9719 (changes requested):
Looks good overall, I have a couple of suggestions

src/reuse/modules/nonUi5/assertion.ts
uaroragit: Added the missing function call statement.

src/reuse/modules/nonUi5/assertion.ts
uaroragit: The purpose behind using getText was to extract the visible text content of an element which strips out the html tags and hidden elements by default.

However, the getHtml method returns the element's html markup i.e. includes the tags as well. 

Link :- https://webdriver.io/docs/api/element/getHTML/

image

uladhsi (approved):

PR #377, opened 2025-05-14
Table select row

DIFSRIP:
Co-authored-by: Arora, Umang <umang.arora+sap@sap.com>

### DIFF
--- a/src/reuse/modules/ui5/table.ts
+++ b/src/reuse/modules/ui5/table.ts
@@ -233,17 +233,12 @@ export class Table {
         }
       };
       if (await ui5.element.isVisible(constructedSelector)) return constructedSelector;
-      // Throw an error if the selector is non of both types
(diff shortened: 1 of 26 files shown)


src/reuse/modules/ui5/table.ts
```diff
@@ -156,40 +157,88 @@ export class Table {
...
 
-    return smartTableSelector;
+      // Check if passed element ID is for a Table
+      constructedSelector = {
+        elementProperties: {
+          metadata: TABLE_METADATA,
+          id: tableSelector
+        }
+      };
+      if (await ui5.element.isVisible(constructedSelector)) return constructedSelector;
```
DIFSRIP, line 235: as discussed in our call, lets switch back to `Promise.any` to not waste time during execution

uaroragit: Pushed the updated file.

PR #395, opened 2025-05-21
refactor: reorganize table operations and add select/open functions

uaroragit:


test/reuse/ui5/table/selectRowByValues.spec.js
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  it("Verification", async function () {
+    const isSelected = await ui5.element.getPropertyValue(checkBoxSelector3("/C_STTA_SalesOrder_WD_20*SalesOrder='500000007'*"), "selected");
+    common.assertion.expectEqual(isSelected, true);
+  });
+});
+
+describe("table - selectRowByValues - demo kit - already selected row remains selected", function () {
+  async function validateChecked() {
```
uladhsi, line 99: This one is also used at `selectRowByIndex`, can we move it into a separate `helper.ts` file, and reuse it from there in both places?

uladhsi (changes requested):
Looks pretty good, I've added some comments regarding improving the code readability

uladhsi (changes requested):
Please see my comments

PR #403, opened 2025-05-27
fix logout: attempt clicking new button first, then click old button

uladhsi:


src/reuse/modules/ui5/navigationBar.ts
```diff
@@ -68,30 +68,43 @@ export class NavigationBar {
...
+    // attempt to click the old user icon
+    try {
+      await scrollAndClickUserIconOld();
+      return;
+    } catch (error) {
+      this.ErrorHandler.logException(error);
     }
 
-    async function clickUserIconOld() {
+    this.ErrorHandler.logException(
```
heraldantony, line 87: Don't think it will hit this code, since `this.ErrorHandler.logException(error);` in line 84 will throw exception if both old and new fail

heraldantony (dismissed):
looks good

src/reuse/modules/ui5/navigationBar.ts
uladhsi: Removed this call and added a warning into the previous catch block

PR #410, opened 2025-06-03
always use new ui5 load function + update timeouts for logout function

uladhsi:


src/reuse/modules/ui5/session.ts
```diff
@@ -346,14 +346,12 @@ export class Session {
           }
         }
       };
-      await ui5.userInteraction.scrollToElement(selector, 0, "end", 500);
```
DIFSRIP, line 349: please clarify with Herald if we can safely remove the scrolls

src/scripts/hooks/utils/Ui5ExtensionMocker.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export class Ui5ExtensionMocker {
+  private static readonly urlRegexps: RegExp[] = [
+    /.*\/xRayControls\/.*/gm,
+    /.*joule.*/gm,
+    /.*walkme.*/gm
+  ];
+
+  private static readonly mockResponse: MockResponse = {
+    payload: "File not found",
```
DIFSRIP, line 17: Maybe something more transparent, like: 
```suggestion
    payload: "File intentionally mocked by Qmate to improve performance",
```

src/scripts/hooks/utils/Ui5ExtensionMocker.ts
```diff
@@ -0,0 +1,30 @@
+interface MockResponse {
+  payload: string;
+  params: {
+    statusCode: number;
+    fetchResponse: boolean;
+  }
+}
+
+export class Ui5ExtensionMocker {
+  private static readonly urlRegexps: RegExp[] = [
```
DIFSRIP, line 10: As suggested by Jan, I would generally support mocking each request specifically. With this current solution we can only respond with the same payload for each request.

@PhyNerd what do you think? You mentioned you have mocked one requests specifically returning something instead of 404?

DIFSRIP (commented):
Please see my comments

src/reuse/modules/ui5/session.ts
uladhsi: reverted this change. Yesterday I saw, that this scroll is necessary on smaller screens
https://github.com/SAP/wdio-qmate-service/pull/410/commits/1031d2b59254ee87c56548bcbc5c31b008fd030b

heraldantony: Logout works for me if I increase the timeout here -
https://github.com/SAP/wdio-qmate-service/blob/ad3378c05e4a90c0694bb1264499805ec0a26659/src/reuse/modules/ui5/navigationBar.ts#L114
The page loads slow for me, may not be an issue for others.
I would suggest that the message be changed here to indicate the call timed out.
https://github.com/SAP/wdio-qmate-service/blob/ad3378c05e4a90c0694bb1264499805ec0a26659/src/reuse/modules/ui5/navigationBar.ts#L115
to 
`timeoutMsg: \`Could not click User Icon in ${+timeout/1000}s\`,`

Same here - https://github.com/SAP/wdio-qmate-service/blob/ad3378c05e4a90c0694bb1264499805ec0a26659/src/reuse/modules/ui5/session.ts#L373
`timeoutMsg: \`Could not click Sign out button in ${+timeout/1000}s\`,`

src/scripts/hooks/before.ts
```diff
@@ -18,4 +19,5 @@ export default async function (capabilities: object[], specs: string[], browser:
   //Add authenticators
   await authenticatorHandler.attachAuthHandling();
 
+  await Ui5ExtensionMocker.mockRequests();
```
DIFSRIP, line 22: we definitely need a config flag to disable mocking in case someone needs these components.

I suggest adding a new config param like:
```ts
qmate: {
  enableUi5ExtensionMocking: true
}
```

- Default is true.
- I would introduce this new `qmate`  options object to the config and in the near future, move all qmate related arguments to the same object for better transparency.

DIFSRIP (changes requested):
Looks good overall. I have one suggestion.

DIFSRIP (changes requested):

PR #411, opened 2025-06-05
add PATCH operation to OData

Templada:


DIFSRIP:
can we proceed with this PR? Do we need to add tests?

cvakiitho (approved):

PR #426, opened 2025-07-01
fix: wild cards did not match across new lines, adding 's-flag'

Templada:
"regression" issue: 
wildcards in selector does not match across new lines/line breaks:

actual value in text property: `"Files in process: 1\n\nCAMT_053.js (ID: 8619)"`
selector: 
`
  const messageSelector = {
          "elementProperties": {
              "metadata": "sap.m.Text",
              "text": "Files in process: 1*"
          }
        };`


PhyNerd (approved):
LGTM

DIFSRIP:
@uladhsi will create a new release for data picker and can include this in the same

PR #430, opened 2025-07-22
fix getUserId

uladhsi:
This PR implements the following:
1. Return null for userId if accessing home directory throws an error

Apart from that, other chores were made:

1. Covered getUserId with tests using vitest
2. Refactored getUserId


PhyNerd (approved):
LGTM thanks!

Issue #450, opened 2025-08-20
[BUG][MEDIUM] - Advanced Data Handling - Exporting Data not exported


ductaily:
**Describe the bug**
I am trying to use the feature „Exporting Data to JSON Files" as described in https://pages.github.tools.sap/sProcurement/qmate/sections/further-concepts/advanced-data-handling/#exporting-data.
I added the export configuration in the config params (e.g, exportData) and use in the test browser.params.export.exportData.
However, after the test runs, no json files were created and I see the error message 

```txt
Unexpected error - no file name for key exportData
```

The bug seem to be in https://github.com/SAP/wdio-qmate-service/blob/5a80102b45c70f794c18a06c49a9e96aac2146e4/src/scripts/dataExchange/dataExchange.ts#L36,  where we are passing the **reference** of `browser.params.export.`
In the following method calls (`copyExportParamsToBrowser`, `copyExportParamsToConfig`), we are overwriting `browser.params.export.*` with `null`, which causes also `exportParams` values also to be null since we are passing a **reference** here and only make shallow copies.

**To Reproduce**

See example created: https://github.tools.sap/CentralInvoices/webdriver-e2e-docker/pull/1210

1. Create configuration:
```js
{
   params: {
     export: {
            MyData: './export/MyData.json'
        },
   // ...
   }
}
```
3. Use the export functionality in your test file.
```js
// example.spec.ts
            browser.config.params.export.MyData = { 'myKey' : 'test' };

```

**Expected behavior**
After running spec, a file `/export/MyData.json` should have been created.

**Screenshots**

**System information:**
 - OS: Mac
 - Browser Chrome - Version 139.0.7258.128
 - Qmate Version 3.11.3, 4.3.4

**Additional context**

ductaily:
@heraldantony Thanks a lot for fixing this bug.

PR #451, opened 2025-08-25
fix bug in export of data to files

heraldantony:
Fix for issue - https://github.com/SAP/wdio-qmate-service/issues/450

### DIFF
--- a/src/scripts/dataExchange/dataExchange.ts
+++ b/src/scripts/dataExchange/dataExchange.ts
@@ -32,7 +32,7 @@ class DataExchange {
 
     // copy filenames, since the user will overwrite browser.params.export with data
     // the filenames are required to write the json data at end of session
-    const exportParams = browser.params.export;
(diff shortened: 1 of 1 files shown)


src/scripts/dataExchange/dataExchange.ts
```diff
@@ -32,7 +32,7 @@ class DataExchange {
 
     // copy filenames, since the user will overwrite browser.params.export with data
     // the filenames are required to write the json data at end of session
-    const exportParams = browser.params.export;
+    const exportParams = { ...config.params.export };
```
uladhsi, line 35: Is there a reason we need to use the spread operator (...) and not just assign the `config.params.export` directly?

uladhsi (changes requested):
Thank you for taking a look into the issue

Just one question from my side, please see the comment

uladhsi (approved):

PR #464, opened 2025-09-08
feat: add new textarea metadata type for ui5 fill function

uladhsi:
Added a new type for metadata of a text area element ("sap.fe.macros.field.TextAreaEx"), which was seen in one of the systems

Also updated types in the package.json file, since otherwise building the qmate-profiles repository's code fails


uaroragit (approved):
Looks good!

PR #468, opened 2025-09-18
refactor/fix: streamline row selection logic in Table class

Templada:
fix for selectRowByIndex() - tables having css checkboxes, instead of sap.m.Checkbox

streamline and reuse private function.

### DIFF
--- a/src/reuse/modules/ui5/table.ts
+++ b/src/reuse/modules/ui5/table.ts
@@ -396,19 +396,8 @@ export class Table {
   async selectRowByIndex(tableSelectorOrId: Ui5Selector | string, index: number) {
     this.vlf.initLog(this.selectRowByIndex);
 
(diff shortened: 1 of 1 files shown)


src/reuse/helper/tableHelper.ts
```diff
@@ -164,13 +164,33 @@ export class TableHelper {
...
+    targetValues: string[],
+    enableHighlighting = true,
+    matchMode: "contains" | "exact" | "wordBoundary" = "contains"
+  ): Promise<string[] | undefined> {
     const matchedRows = rows.filter((row: any) => {
       const cells = row.getCells();
       return targetValues.every((val) =>
         cells.some((cell: any) => {
           const domRef = cell.getDomRef();
-          return domRef && domRef.innerText && domRef.innerText.includes(val);
```
Templada, line 173: got rid of this due to false positives

DIFSRIP (approved):
LGTM

PR #536, opened 2026-04-02
Fix/ui5 table open item

imaroz-lx:
Append table open items methods with two ways how to perform open action.
Now it tries to click on arrow icon at first and hit direct click if cannot find one.

- Add right arrow click functionality to open row in openItemByIndex()
- Update openItemByIndex() tests cases
- Add right arrow click functionality to open row via openItemByValues()
- Update openItemByValues() tests cases

Issue details: In some cases table items can contain interactive elements (like inputs) and when they appears in the way qmate clicks on this elements rather than tries to actually open the item.


src/reuse/modules/ui5/table.ts
```diff
@@ -578,19 +609,44 @@ export class Table {
...
+        elementProperties: { 
+          metadata: "sap.ui.core.Icon", 
+          src: "sap-icon://slim-arrow-right"
+        },
+        parentProperties: rowSelector.elementProperties
+      }
+      await ui5.userInteraction.click(rowArrowIconSelector);
+    }
+    if(triggerOption === "byEvent") {
+      const elem = await ui5.element.getDisplayed(rowSelector, index);
```
uaroragit: This could introduce a bug — rowSelector is already derived using rowSelectors[index], so passing index again results in double indexing.

This can lead to incorrect element selection or flaky failures when only one element exists.

src/reuse/modules/ui5/table.ts
```diff
@@ -544,11 +545,40 @@ export class Table {
...
+      const rowArrowIconSelector: Ui5Selector = {
+        elementProperties: { 
+          metadata: "sap.ui.core.Icon", 
+          src: "sap-icon://slim-arrow-right"
+        },
+        parentProperties: rowSelector.elementProperties
+      }
+      await ui5.userInteraction.click(rowArrowIconSelector);
+    }
+    if(triggerOption === "byEvent") {
```
uaroragit: These options are mutually exclusive. Using separate 'if' evaluates all conditions, which could run multiple blocks if logic changes (example: adding overlapping conditions, or extra checks etc). So I would suggest using 'else if' to ensure only one path executes and clarifies intent.

test/reuse/ui5/table/openItemByIndex.spec.js
```diff
@@ -15,7 +15,7 @@ const goButtonSelector = {
   }
```
uaroragit: There might be changes required to the individual methods and hence subsequently to these unit tests. So I would prefer reviewing them once above suggestions are addressed. Thanks!

uaroragit (changes requested):
Thank you for the changes. Please see my recommendations.

DIFSRIP:
@imaroz-lx  @uaroragit can we merge this PR?

src/reuse/modules/ui5/table.ts
```diff
@@ -588,12 +604,24 @@ export class Table {
...
+        metadata: "sap.ui.core.Icon", 
+        src: "sap-icon://slim-arrow-right"
+      },
+      parentProperties: rowSelector.elementProperties
+    }
+
+    if(await ui5.element.isVisible(rowArrowIconSelector)) {
+      await ui5.userInteraction.click(rowArrowIconSelector);
+    }
+    else {
```
DIFSRIP: This seems to be very similar to the added code above. Should we move it into a private function and reuse it on both places? Or maybe at least the selector as global const of the file?

DIFSRIP (changes requested):

src/reuse/modules/ui5/table.ts
imaroz-lx: Done. Moved to the private method

src/reuse/modules/ui5/table.ts
```diff
@@ -1002,5 +980,21 @@ export class Table {
         throw new Error("No selectable element found for the row.");
     }
   }
+
+  private async _clickRowArrowIcon(rowSelector: Ui5Selector): Promise<boolean> {
```
DIFSRIP: Now we are mixing up the thing on 2 places. 
I suggest to rework the helper to perform both clicks internally, no return value or parent if required.

Do function should either click on or the other, so make sure to rename it to smth like: `_clickRowOrArrowIcon`

DIFSRIP (changes requested):

src/reuse/modules/ui5/table.ts
imaroz-lx: Done

DIFSRIP (approved):

PR #541, opened 2026-04-08
fix: remove clearValue method and apply browser execute

imaroz-lx:
Yet another fix on ui5.userInteraction.clear method.
- Replace wdio clearValue() method with browser script
- Switch code parts placement to reduce code by removing elem.click() to leave focuse
- Add descriptive comment on why builtin clearValue should not be used in this method

Issue details: Some of the inputs are placed on focus sensitive elements (like table column filters popovers) that hides right after clearValue() was called. And if that element change it state only when "Enter" is pressed then after popover closes changes are not apply.
Screenshot 2026-04-08 at 19 46 10


heraldantony (approved):
tested, works fine.

DIFSRIP (approved):

PR #544, opened 2026-04-14
feat: enhance hex validation to support base64 encoded hex strings

DIFSRIP:
The previous PR https://github.com/SAP/wdio-qmate-service/pull/518 introduced a regression for base64 decoded hex values. 

This PR addresses the issue and adds support for both formats.


heraldantony (changes requested):
I tested, and it works (for base64 encoded strings).

AI generated review:

__1. False positive risk with `_isBase64EncodedHex`__ `Buffer.from()` with `"base64"` encoding is very lenient — it silently ignores invalid characters rather than throwing. This means the try/catch will almost never catch. For example, `Buffer.from("CAFE", "base64")` succeeds and decodes to bytes `08 00 44`, which in utf8 could coincidentally match hex patterns.

For short strings, there's a non-trivial risk a plain-text value could base64-decode to something that looks like hex. Consider adding a __minimum decoded length check__ (RSA-encrypted values are typically ≥128 chars):

```ts
return /^[0-9a-fA-F]+$/.test(decoded) && decoded.length % 2 === 0 && decoded.length >= 64;
```

__2. Option merge order is correct but subtle__

```ts
const effectiveOptions = isBase64Hex ? { useBase64Input: true, ...options } : options;
```

Since `...options` spreads __after__ `useBase64Input: true`, user-provided `useBase64Input: false` would override the auto-detection. This is the right behavior (explicit user intent wins), but a brief inline comment would help future readers:

```ts
// User-provided options override auto-detected values
const effectiveOptions = isBase64Hex ? { useBase64Input: true, ...options } : options;
```

__3. `_alreadyDecryptedData` grows unboundedly__ In `getSecureData`, `dataIdentifier` is pushed to `_alreadyDecryptedData` __every call__ (line 87), even if already present. Over many calls to the same file, the array grows. Consider using a `Set` instead:

```ts
private _alreadyDecryptedData: Set<string> = new Set();
// ...
if (!this._alreadyDecryptedData.has(dataIdentifier) && privateKeyFound) { ... }
this._alreadyDecryptedData.add(dataIdentifier);  // Set ignores duplicates
```

*(Note: this pre-dates this PR but worth mentioning)*

---

### Test Coverage ✅
Not that important since we already handle base64 string decryption in decrypt.base64.secure.json

__Missing test (nice-to-have):__ A test with a mixed secure file containing both plain hex AND base64 hex values in the same object, verifying both are decrypted correctly in one `getSecureData` call.

src/reuse/modules/util/data.ts
```diff
@@ -152,18 +152,37 @@ export class Data {
...
       if (typeof data[key] === "object" && !Array.isArray(data[key])) {
         data[key] = this._decryptRecursively(data[key], options);
-      } else if ((typeof data[key] === "string" && this._isHex(data[key])) || Array.isArray(data[key])) {
+      } else if (Array.isArray(data[key])) {
         data[key] = global.util.data.decrypt(data[key], options);
+      } else if (typeof data[key] === "string") {
+        const isPlainHex = this._isHex(data[key]);
+        const isBase64Hex = !isPlainHex && this._isBase64EncodedHex(data[key]);
+
+        if (isPlainHex || isBase64Hex) {
```
heraldantony, line 165: Probably better to add the check for string length here, since we don't want to decrypt either hex or base64 strings <= 64 chars long, since they are not encrypted.
Or add an additional check in _isHex function.

```JavaScript
if ( data[key].length >= 64 && (isPlainHex || isBase64Hex)) {
```

DIFSRIP: Thanks for the feedback, applied your points. One small deviation on the length check: 
instead of adding it at the call site, I moved it into `_isHex `directly `(str.length >= 64`). 
This avoids the operator-precedence ambiguity in the suggested snippet (`data[key].length >= 64 && isPlainHex || isBase64Hex` evaluates the base64 branch without the length guard). 
As a bonus, `_isBase64EncodedHex `was refactored to delegate to `_isHex(decoded)`, so both detectors share a single validation path, the length threshold only lives in one place.

heraldantony (approved):
Tested, works fine. Fix looks good.

PR #546, opened 2026-04-20
fix: click resolves foreground element when multiple dialogs share the same selector

uaroragit:
When two dialogs are open simultaneously (e.g., a background dialog and a foreground modal), 'ui5.userInteraction.click' fails with "Element not clickable after 30s" even though the correct element was perfectly clickable.

Root cause: 'browser.uiControls()' returns all matching UI5 controls across both dialogs. The element at index 0 was always the background dialog's element — which is covered by the foreground modal overlay and therefore not clickable. The 'waitUntil' loop polled the same non-clickable element for the full timeout duration without ever considering the foreground element.

### DIFF
--- a/src/reuse/modules/ui5/userInteraction.ts
+++ b/src/reuse/modules/ui5/userInteraction.ts
@@ -22,6 +22,7 @@ export class UserInteraction {
   private static readonly TEXTAREA_MACROS_METADATA: Ui5ControlMetadata = "sap.fe.macros.field.TextAreaEx";
   private static readonly SUPPORTED_TEXTAREA_METADATA: Array<Ui5ControlMetadata> = [UserInteraction.TEXTAREA_METADATA, UserInteraction.TEXTAREA_MACROS_METADATA];
   private static readonly SELECT_DEPRECATION_MESSAGE: string = "This function is deprecated, please use the generic 'ui5.userInteraction.select' function instead.";
+  private static readonly OVERLAY_CHECK_TIMEOUT = 5000;
 
   // =================================== CLICK ===================================
   /**
@@ -35,22 +36,8 @@ export class UserInteraction {
    */
   async click(selector: any, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT) {
     const vl = this.vlf.initLog(this.click);
-    let elem = null;
-    const timeoutMsg = `Element not clickable after ${+timeout / 1000}s`;
+    const elem = await this._waitForClickable(selector, index, timeout);
     try {
-      await browser.waitUntil(
-        async function () {
-          elem = await ui5.element.getDisplayed(selector, index, timeout);
-          if (!elem) return false;
-          return elem.isClickable();
-        },
-        { timeout, timeoutMsg }
-      );
-    } catch (error: any) {
-      elem = await this._fallbackOnOverlay(error, timeoutMsg, selector, index);
-    }
-    try {
-      // @ts-ignore
       await elem.click();
     } catch (error) {
       // @ts-ignore
@@ -85,22 +72,8 @@ export class UserInteraction {
    */
   async doubleClick(selector: any, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT) {
     const vl = this.vlf.initLog(this.doubleClick);
-    let elem = null;
-    const timeoutMsg = `Element not clickable after ${+timeout / 1000}s`;
+    const elem = await this._waitForClickable(selector, index, timeout);
     try {
-      await browser.waitUntil(
-        async function () {
-          elem = await ui5.element.getDisplayed(selector, index, timeout);
-          if (!elem) return false;
-          return elem.isClickable();
-        },
-        { timeout, timeoutMsg }
-      );
-    } catch (error: any) {
-      elem = await this._fallbackOnOverlay(error, timeoutMsg, selector, index);
-    }
-    try {
-      // @ts-ignore
       await elem.doubleClick();
     } catch (error) {
       // @ts-ignore
@@ -120,22 +93,8 @@ export class UserInteraction {
    */
   async rightClick(selector: any, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT) {
     const vl = this.vlf.initLog(this.rightClick);
-    let elem = null;
-    const timeoutMsg = `Element not clickable after ${+timeout / 1000}s`;
+    const elem = await this._waitForClickable(selector, index, timeout);
     try {
-      await browser.waitUntil(
-        async function () {
-          elem = await ui5.element.getDisplayed(selector, index, timeout);
-          if (!elem) return false;
-          return elem.isClickable();
-        },
-        { timeout, timeoutMsg }
-      );
-    } catch (error: any) {
-      elem = await this._fallbackOnOverlay(error, timeoutMsg, selector, index);
-    }
-    try {
-      // @ts-ignore
       await elem.click({ button: "right" });
     } catch (error) {
       // @ts-ignore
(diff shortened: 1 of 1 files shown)


src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -730,6 +693,47 @@ export class UserInteraction {
   }
 
   // =================================== HELPER ===================================
+  private async _waitForClickable(selector: any, index: number, timeout: number): Promise<Element> {
```
DIFSRIP: I suggest to rename the function as the current name sounds like a void return type.

--> something like:  `_getClickableElement `

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -730,6 +693,47 @@ export class UserInteraction {
...
+    const poll = async () => {
+      elem = await ui5.element.getDisplayed(selector, index, timeout);
+      return !!(elem && await elem.isClickable());
+    };
+
+    try {
+      await browser.waitUntil(poll, { timeout: firstPhaseTimeout, timeoutMsg: firstPhaseMsg });
+      return elem!;
+    } catch (e: any) {
+      if (e?.message !== "quick") throw e;
```
DIFSRIP: I was initially confused by the implementation and thought the below code will never be reached.

I guess we can improve that, by adding a comment here, explaining what "quick" means and when the function will only throw.

DIFSRIP: If the error message is not "quick"
→ throw → function exits

If the message is "quick"
→ swallow the error
→ execution continues after the try/catch

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -730,6 +693,47 @@ export class UserInteraction {
...
+    }
+
+    // Check once, outside the polling loop, whether a UI5 block layer (class "sapUiBLy") is
+    // physically covering the center of this specific element.
+    const isBlocked: boolean = await browser.execute((el: HTMLElement) => {
```
DIFSRIP: I think this could be converted into an inner function and moved at the end of the file to improve the reading flow.

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -730,6 +693,47 @@ export class UserInteraction {
...
+      const rect = el.getBoundingClientRect();
+      const topElem = document.elementFromPoint(rect.left + rect.width / 2, rect.top + rect.height / 2);
+      return topElem?.classList?.contains("sapUiBLy") ?? false;
+    }, elem as any);
+    if (isBlocked) {
```
DIFSRIP: Maybe add a comment before, to explain when this code will be executed. Something like:

```ts
// Check if it is blocked if the element does exist but it just isn’t clickable yet
```

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -730,6 +693,47 @@ export class UserInteraction {
...
+      for (let i = index + 1; ; i++) {
+        try {
+          const candidate = await ui5.element.getDisplayed(selector, i, 5000);
```
DIFSRIP, line 769: Please add a comment when the loop will terminate. On the first look, it seemed infinite. 

```ts
// loop exits when getDisplayed fails -> there is no further element present/no element at that index
```

DIFSRIP (changes requested):
Thanks for the implementation. After walking through it in detail, the logic itself makes sense, but I was initially confused by the control flow and thought some parts of the code would never be reached. I think a few small readability improvements could make the intent much clearer and reduce the mental load for future readers.

***

#### 1. Clarify the `"quick"` timeout behavior

At first glance, it looks like the code after the first `try/catch` would never execute. The fact that execution can continue depends on the special `"quick"` timeout message, which isn’t obvious when reading the code.

It would help to add a comment explaining this explicitly, e.g.:

```ts
// If the error message is not "quick", we fail fast and exit the function.
// If the message *is* "quick", only the initial short wait expired.
// This means the element exists but is not clickable yet, and we
// intentionally continue with additional checks instead of throwing.
```

Without this context, it’s easy to assume the function always throws here.

***

#### 2. Explain when the `isBlocked` logic is executed

The `isBlocked` check is only reached under a very specific combination of conditions, which isn’t obvious on a first read.

A short comment before this block would help, for example:

```ts
// Reached only when:
// - the element exists and is displayed
// - it is not clickable yet
// - the initial "quick" wait timed out
// Now check whether the element is physically blocked by a UI5 block layer.
```

***

#### 3. Convert the overlay check into an inner helper function

The inline `browser.execute` block interrupts the reading flow and mixes low‑level DOM inspection into the middle of the main wait logic.

I think converting this into a small **inner helper function** (e.g. `_isBlockedByUi5Overlay(elem)`) would improve readability significantly and allow the core control flow to be read top‑to‑bottom, with the overlay‑specific logic moved to the end of the function where it doesn’t disturb the main logic.

***

#### 4. Document the termination condition of the loop

The `for (;;)` loop initially looks infinite. While it is intentionally unbounded, it relies on `getDisplayed` throwing once no further elements exist.

A short comment would make this much clearer, e.g.:

```ts
// Try alternative elements with the same selector.
// The loop terminates when getDisplayed throws, which indicates
// that no further element exists at the given index.
```

***

## AI Generated improvement (to be verified)

```ts
private async _getClickableElement(
  selector: any,
  index: number,
  timeout: number
): Promise<Element> {

  let elem: Element | null = null;
  const timeoutMsg = `Element not clickable after ${timeout / 1000}s`;

  // Phase 1: short wait used to quickly detect overlay situations
  const firstPhaseTimeout = Math.min(timeout, UserInteraction.OVERLAY_CHECK_TIMEOUT);
  const isQuickPhase = firstPhaseTimeout < timeout;

  const QUICK_TIMEOUT_MSG = "__QUICK_TIMEOUT__";

  const poll = async () => {
    elem = await ui5.element.getDisplayed(selector, index, timeout);
    return !!(elem && await elem.isClickable());
  };

  try {
    // If the element becomes clickable during the quick phase,
    // return immediately.
    await browser.waitUntil(poll, {
      timeout: firstPhaseTimeout,
      timeoutMsg: QUICK_TIMEOUT_MSG
    });
    return elem!;
  } catch (e: any) {

    // If the error message is not the "quick" timeout, we fail fast
    // and exit the function.
    //
    // If it *is* the "quick" timeout, only the initial short wait expired.
    // This means the element exists but is not clickable yet, and we
    // intentionally continue with additional checks instead of throwing.
    if (!isQuickPhase || e?.message !== QUICK_TIMEOUT_MSG) {
      throw e;
    }
  }

  // Reached only when:
  // - the element exists and is displayed
  // - it is not clickable yet
  // - the initial "quick" wait timed out
  // Now check whether the element is physically blocked by a UI5 block layer.
  if (await isBlockedByUi5Overlay(elem!)) {

    // Try alternative elements with the same selector.
    // The loop terminates when getDisplayed throws, which indicates
    // that no further element exists at the given index.
    for (let i = index + 1; ; i++) {
      try {
        const candidate = await ui5.element.getDisplayed(selector, i, 5000);
        if (await candidate.isClickable()) {
          return candidate;
        }
      } catch {
        throw new Error(timeoutMsg);
      }
    }
  }

  // Element is not blocked — continue waiting for the remaining time
  await browser.waitUntil(poll, {
    timeout: timeout - firstPhaseTimeout,
    timeoutMsg
  });

  return elem!;

  /**
   * Checks whether the given element is physically covered by a UI5 block layer
   * (sapUiBLy) at its center position.
   */
  async function isBlockedByUi5Overlay(element: Element): Promise<boolean> {
    return browser.execute((el: HTMLElement) => {
      const rect = el.getBoundingClientRect();
      const topElem = document.elementFromPoint(
        rect.left + rect.width / 2,
        rect.top + rect.height / 2
      );
      return topElem?.classList?.contains("sapUiBLy") ?? false;
    }, element as any);
  }
}
```

src/reuse/modules/ui5/userInteraction.ts
uaroragit: Renamed to '_getClickableElement'

src/reuse/modules/ui5/userInteraction.ts
uaroragit: Added suggested comments.

src/reuse/modules/ui5/userInteraction.ts
uaroragit: Moved the logic to check presence of overlay into inner function 'isBlockedByUi5Overlay'

src/reuse/modules/ui5/userInteraction.ts
uaroragit: Slightly modified comment added ("// Element was found but remained non-clickable throughout the first phase") at top of two similar intended comments, so that overall it makes sense. 

Hope it's fine.

src/reuse/modules/ui5/userInteraction.ts
uaroragit: Added comment at top of the loop

uaroragit:
Hi @DIFSRIP 
Adopted all the readability improvements — isQuickPhase, QUICK_TIMEOUT_MSG, JSDoc, and the formatting changes.

One small adjustment on line 745:
timeoutMsg: isQuickPhase ? QUICK_TIMEOUT_MSG : timeoutMsg

When timeout ≤ 5000, isQuickPhase is false and both phases collapse into one. If waitUntil then times out, the catch re-throws e (because !isQuickPhase is true), exposing "__QUICK_TIMEOUT__" to the caller instead of "Element not clickable after 3s".

This can happen when a short timeout is passed explicitly (click(selector, 0, 3000)) or via QMATE_CUSTOM_TIMEOUT=3000. Passing timeoutMsg for the non-quick-phase case keeps QUICK_TIMEOUT_MSG as an internal marker and the user-facing error message meaningful.

DIFSRIP (approved):

PR #548, opened 2026-04-22
Feat/ui5 user interaction open value help

imaroz-lx:
Update openF4Help method by adding openValueHelp and deprecating old one.

- Add ui5 user interaction openValueHelp method
- Add tests for openValueHelp method
- Deprecate openF4Help method

### DIFF
--- a/src/reuse/modules/ui5/userInteraction.ts
+++ b/src/reuse/modules/ui5/userInteraction.ts
@@ -696,6 +696,7 @@ export class UserInteraction {
    * @param {Object} selector - The selector describing the element.
    * @param {Number} [index=0] - The index of the selector (in case there are more than one elements visible at the same time).
    * @param {Number} [timeout=30000] - The timeout to wait (ms).
+   * @param {Boolean} useF4Key - Specifies if the value help is opened by pressing the F4-key or via the value help icon click.
    * @example 
    * ```
    * const selector = {
@@ -706,14 +707,20 @@ export class UserInteraction {
    *     id: "element-id-vhi"
    *   }
    * }
-   * await ui5.userInteraction.openValueHelp(selector, 0, 30000);
+   * await ui5.userInteraction.openValueHelp(selector);
+   * await ui5.userInteraction.openValueHelp(selector, 0, 30000, true);
    * ```
    */
-  async openValueHelp(selector: Ui5Selector, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT) {
+  async openValueHelp(selector: Ui5Selector, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT, useF4Key: boolean = false) {
     const vl = this.vlf.initLog(this.openValueHelp);
-    const id = await ui5.element.getId(selector, index, timeout);
-    const button = await nonUi5.element.getByCss(`[id='${id}-vhi']`, 0, timeout);
-    await button.click();
+    if (useF4Key === true) {
+      await ui5.userInteraction.click(selector, index, timeout);
+      await common.userInteraction.pressF4();
+    } else {
+      const id = await ui5.element.getId(selector, index, timeout);
+      const button = await nonUi5.element.getByCss(`[id='${id}-vhi']`, 0, timeout);
+      await button.click();
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -678,6 +679,7 @@ export class UserInteraction {
    */
   async openF4Help(selector: any, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT, useF4Key = true) {
```
DIFSRIP, line 681: We can remove the annotations to hide it in the API docs.

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -688,6 +690,33 @@ export class UserInteraction {
     }
```
DIFSRIP: If we correctly deprecate the function with backwards compatibility of the new one, we should internally call the new one instead of the parallel logic. 

I guess for that, we have to support the `useF4Key ` arg as well and I think it is worth doing that?

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -688,6 +690,33 @@ export class UserInteraction {
...
+   *   }
+   * }
+   * await ui5.userInteraction.openValueHelp(selector, 0, 30000);
+   * ```
+   */
+  async openValueHelp(selector: Ui5Selector, index = 0, timeout: number = parseFloat(process.env.QMATE_CUSTOM_TIMEOUT!) || GLOBAL_DEFAULT_WAIT_TIMEOUT) {
+    const vl = this.vlf.initLog(this.openValueHelp);
+    const id = await ui5.element.getId(selector, index, timeout);
+    const button = await nonUi5.element.getByCss(`[id='${id}-vhi']`, 0, timeout);
+    await button.click();
```
DIFSRIP: If we support the former `useF4Key ` arg, we need to update the logic here and only click the icon if its is false (default)

test/reuse/ui5/userInteraction/openValueHelp.spec.js
```diff
@@ -0,0 +1,55 @@
+const { BASE_URL } = require("../../../../src/reuse/constants.ts");
```
DIFSRIP, line 1: Just a note: let's make sure, the old test cases are still running when we call the new logic internally for `openF4Help`

DIFSRIP (changes requested):
Thanks for the refactoring. Please see my comments.

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Done

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Applied new logic as part of the logic in the old method. Is it enough or the key pressing logic need to be implemented as well?

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: See comment above

test/reuse/ui5/userInteraction/openValueHelp.spec.js
imaroz-lx: Don't really understand the issue here. The old tests are leaved untouched in openF4Help.spec.ts

src/reuse/modules/ui5/userInteraction.ts
DIFSRIP: I would also implement the key logic as part of the new and here simply call the new one + the deprecation warning since you user could also use the key here.

src/reuse/modules/ui5/userInteraction.ts
DIFSRIP: I still see them. We need to remove them, so the function will no longer be exposed

src/reuse/modules/ui5/userInteraction.ts
DIFSRIP: let's support that logic

test/reuse/ui5/userInteraction/openValueHelp.spec.js
DIFSRIP: yes, no action required, just to doublecheck we do not remove the older tests for now

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Done

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Done

test/reuse/ui5/userInteraction/openValueHelp.spec.js
imaroz-lx: Done. Also added same tests to openValue method to ensure test coverage in case old method with test would be deleted

DIFSRIP (approved):

PR #557, opened 2026-05-01
Fix/invisible elements jamming

imaroz-lx:
Some special elements can breaks child/parent selector filtering. See [details](https://github.tools.sap/sProcurement/qmate/issues/1123)

### DIFF
--- a/src/scripts/locators/qmateLocatorSrc/constants/IgnoredMetadataList.ts
+++ b/src/scripts/locators/qmateLocatorSrc/constants/IgnoredMetadataList.ts
@@ -1,5 +1,6 @@
-// List of sap elements metadata's that exists but actually not presented in the UI tree.
-// Elements with this metadata's should be ignored while filtering valid controls
+// These UI5 controls exist in the control registry but sit outside the standard UI5 control tree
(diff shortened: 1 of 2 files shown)


src/scripts/locators/qmateLocatorSrc/constants/IgnoredMetadataList.ts
```diff
@@ -0,0 +1,5 @@
+// List of sap elements metadata's that exists but actually not presented in the UI tree.
+// Elements with this metadata's should be ignored while filtering valid controls
```
DIFSRIP: To add some context: 
```suggestion
// These UI5 controls exist in the control registry but sit outside the standard UI5 control tree
// (e.g. sap.ui.core.HTML renders raw HTML and has no UI5 parent/child relationships). 
// Including them during DOM traversal breaks parent and child filtering, so they are skipped and their DOM children are inspected directly instead.
```

DIFSRIP (changes requested):
Looks good, just a minor comment adjustment

DIFSRIP (approved):

PR #559, opened 2026-05-06
Fix/ui5 user interaction click block

imaroz-lx:
Click method not always correctly resolve where is the unblocked element placed. See [details](https://github.tools.sap/orgs/sProcurement/projects/17/views/2?pane=issue&itemId=304743&issue=sProcurement%7Cqmate%7C1130)

### DIFF
--- a/src/reuse/modules/ui5/userInteraction.ts
+++ b/src/reuse/modules/ui5/userInteraction.ts
@@ -5,7 +5,7 @@ import { VerboseLoggerFactory } from "../../helper/verboseLogger";
 import { AlignmentOptions, AlignmentValues } from "../types";
 import ErrorHandler from "../../helper/errorHandler";
 import { GLOBAL_DEFAULT_WAIT_INTERVAL, GLOBAL_DEFAULT_WAIT_TIMEOUT } from "../constants";
-import { Ui5ControlMetadata, Ui5Selector } from "./types/ui5.types";
+import { CssSelector, Ui5ControlMetadata, Ui5Selector } from "./types/ui5.types";
 import { KeyCodes } from "../common/constants/userInteraction.constants";
 
 /**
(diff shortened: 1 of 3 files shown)


src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -756,24 +764,23 @@ export class UserInteraction {
...
+    if(isBlockActive) {
+      const elems = await ui5.element.getAllDisplayed(selector, timeout);
+      const minZIndex = await resolveZIndex(await nonUi5.element.getByCss(blockLayerSelector));
+      const unblockedElems: Array<Element> = [];
+      for (const elem of elems) {
+          // Element count as non-blocked only if it arranged with or above than block layer.
+          if(await resolveZIndex(elem) >= minZIndex) 
+            unblockedElems.push(elem);
       }
+      if (index >= elems.length || index < 0) this.ErrorHandler.logException(new Error(), indexOutOfBoundsMsg);
```
DIFSRIP: -  If all elements are blocked, unblockedElems is empty, elem becomes undefined, and the subsequent waitUntil fails with a cryptic error rather than a meaningful message.
- The check is against elems.length (all elements) but the assignment uses unblockedElems. If there are 3 elements total but only 1 is unblocked, index = 1 passes the check but unblockedElems[1] is undefined. Should be index >= unblockedElems.length.

```suggestion
    if (unblockedElems.length === 0) throw new Error(timeoutMsg);
    if (index >= unblockedElems.length || index < 0) this.ErrorHandler.logException(new Error(), indexOutOfBoundsMsg);
```

DIFSRIP (changes requested):
Thanks, the new approach seems more reliable!

I commented on a potential bug below.
Also make sure to format the files with prettier

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Good catch. Fixed and updated with more precise error.

DIFSRIP (approved):

DIFSRIP (approved):

PR #560, opened 2026-05-07
style: apply prettier to the project

imaroz-lx:


DIFSRIP (approved):
No manual review done, only style changes.

PR #569, opened 2026-06-01
Fix refactor ui5 click

imaroz-lx:
Whole _getClickableElement private method refactor. Much more simplified and solve existing problems. Here is [details](https://github.tools.sap/sProcurement/qmate/issues/1167)

### DIFF
--- a/src/reuse/modules/ui5/userInteraction.ts
+++ b/src/reuse/modules/ui5/userInteraction.ts
@@ -22,7 +22,6 @@ export class UserInteraction {
   private static readonly TEXTAREA_MACROS_METADATA: Ui5ControlMetadata = "sap.fe.macros.field.TextAreaEx";
   private static readonly SUPPORTED_TEXTAREA_METADATA: Array<Ui5ControlMetadata> = [UserInteraction.TEXTAREA_METADATA, UserInteraction.TEXTAREA_MACROS_METADATA];
   private static readonly SELECT_DEPRECATION_MESSAGE: string = "This function is deprecated, please use the generic 'ui5.userInteraction.select' function instead.";
-  private static readonly OVERLAY_CHECK_TIMEOUT = 5000;
   private static readonly OPENF4HELP_DEPRECATION_MESSAGE: string = "This function is deprecated, please use the generic 'ui5.userInteraction.openValueHelp' function instead.";
 
   // =================================== CLICK ===================================
@@ -728,38 +727,53 @@ export class UserInteraction {
 
   // =================================== HELPER ===================================
   private async _getClickableElement(selector: any, index: number, timeout: number): Promise<Element> {
-    let elem: Element | null = null;
+    let elems: Array<Element> | null = null;
     const nonClickableMsg = `Element is not clickable after ${timeout / 1000}s`;
-    const timeoutMsg = `Not found elements with selector: ${JSON.stringify(selector)}`;
+    const indexOutOfBoundsMsg = `Index is out of bounds. No elements with index: ${index}`;
 
-    await browser.waitUntil(
-      async () => {
-        let elems: Array<Element> = [];
-        await browser.waitUntil(
-          async () => {
-            elems = await ui5.element.getAllDisplayed(selector, timeout);
-            return elems.length !== 0;
-          },
-          {
-            timeout: timeout,
-            interval: GLOBAL_DEFAULT_WAIT_INTERVAL,
-            timeoutMsg: timeoutMsg
+    let errorMsg = "Unexpected error";
+    try {
+      await browser.waitUntil(
+        async () => {
+          try {
+            // Handle dialogs overlays
+            const lastOpenedDialog = await this._getLastOpenedDialog();
+            if (lastOpenedDialog) {
+              elems = await lastOpenedDialog.uiControls(selector, timeout);
+            } else {
+              elems = await ui5.element.getAllDisplayed(selector, timeout);
+            }
+          } catch (e) {
+            return ((errorMsg = (e as Error).message), false);
           }
-        );
-        // Non-clickable elements are not counts while indexing
-        const isClickableFlags = await Promise.all(elems.map(async (e) => e.isClickable()));
-        elem = elems.filter((_, i) => isClickableFlags[i])[index];
 
-        return !!elem;
-      },
-      {
-        timeout: timeout,
-        interval: GLOBAL_DEFAULT_WAIT_INTERVAL,
-        timeoutMsg: nonClickableMsg
-      }
-    );
+          if (index >= elems!.length) return ((errorMsg = indexOutOfBoundsMsg), false);
+          if (!(await elems[index].isClickable())) return ((errorMsg = nonClickableMsg), false);
+          return true;
+        },
+        {
+          timeout: timeout,
+          interval: GLOBAL_DEFAULT_WAIT_INTERVAL
+        }
+      );
+    } catch (e) {
+      this.ErrorHandler.logException(new Error(), errorMsg);
+    }
 
-    return elem!;
+    return elems![index];
+  }
+
+  private async _getLastOpenedDialog(): Promise<Element | undefined> {
+    const vl = this.vlf.initLog(this._getLastOpenedDialog);
+    const lastOpenedDialogId = await browser.execute(() => {
+      const staticAreaDomElems = [...sap.ui.getCore().getStaticAreaRef().children];
+      const sapElems = staticAreaDomElems.flatMap((elem) => sap.ui.getCore().byId(elem.id) ?? []);
+      const dialogs = sapElems.filter((popup) => popup.isA("sap.m.Dialog"));
+      return dialogs.splice(-1)[0]?.getId();
+    });
+    if (!lastOpenedDialogId) return undefined;
+    vl.log(`Found opened dialog with id: ${lastOpenedDialogId}`);
+    return await nonUi5.element.getById(lastOpenedDialogId);
   }
 
   private async _verifyTabSwitch(selector: any): Promise<boolean> {
--- a/src/reuse/modules/ui5/userInteraction.ts
+++ b/src/reuse/modules/ui5/userInteraction.ts
@@ -729,71 +729,36 @@ export class UserInteraction {
   // =================================== HELPER ===================================
   private async _getClickableElement(selector: any, index: number, timeout: number): Promise<Element> {
     let elem: Element | null = null;
-    const allBlockedMsg = `No clickable elements found with selector: ${JSON.stringify(selector)}`;
-    const indexOutOfBoundsMsg = `Index is out of bounds. No elements with index: ${index}`
-    const timeoutMsg = `Element not clickable after ${timeout / 1000}s`;
-    const QUICK_TIMEOUT_MSG = "__QUICK_TIMEOUT__";
-    
-    const firstPhaseTimeout = Math.min(timeout, UserInteraction.OVERLAY_CHECK_TIMEOUT);
-    const isQuickPhase = firstPhaseTimeout < timeout;
+    const nonClickableMsg = `Element is not clickable after ${timeout / 1000}s`;
+    const timeoutMsg = `Not found elements with selector: ${JSON.stringify(selector)}`;
 
-    const resolveZIndex = async (elem: Element): Promise<number> => {
-      const zIndex: string = await nonUi5.element.getCssPropertyValue(elem, "z-index");
-      if (zIndex === "auto") {
-        const parentElement = await elem.parentElement();
-        return (parentElement ? await resolveZIndex(parentElement) : 0);
+    await browser.waitUntil(
+      async () => {
+        let elems: Array<Element> = [];
+        await browser.waitUntil(
+          async () => {
+            elems = await ui5.element.getAllDisplayed(selector, timeout);
+            return elems.length !== 0;
+          },
+          {
+            timeout: timeout,
+            interval: GLOBAL_DEFAULT_WAIT_INTERVAL,
+            timeoutMsg: timeoutMsg
+          }
+        );
+        // Non-clickable elements are not counts while indexing
+        const isClickableFlags = await Promise.all(elems.map(async (e) => e.isClickable()));
+        elem = elems.filter((_, i) => isClickableFlags[i])[index];
+
+        return !!elem;
+      },
+      {
+        timeout: timeout,
+        interval: GLOBAL_DEFAULT_WAIT_INTERVAL,
+        timeoutMsg: nonClickableMsg
       }
-      return parseInt(zIndex);
-    }
+    );
 
-    const poll = async () => {
-      elem = await ui5.element.getDisplayed(selector, index, timeout);
-      return !!(elem && (await elem.isClickable()));
-    };
-
-    try {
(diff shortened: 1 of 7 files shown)


src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -729,71 +729,36 @@ export class UserInteraction {
...
+    await browser.waitUntil(
+      async () => {
+        let elems: Array<Element> = [];
+        await browser.waitUntil(
+          async () => {
+            elems = await ui5.element.getAllDisplayed(selector, timeout);
+            return elems.length !== 0;
+          },
+          {
+            timeout: timeout,
```
DIFSRIP: **[Required]** The inner `waitUntil` is given the same full `timeout` as the outer loop (`timeout: timeout` here). This causes two problems:

1. **Double-wait:** The outer loop's predicate blocks until the inner call resolves. When no elements are found, the inner call runs for the full `timeout` ms before throwing — by which point the outer clock has also elapsed. Worst case: `2 × timeout` total wait.
2. **Wrong error message:** The inner times out first and throws `"Not found elements with selector:..."` before the outer ever checks its own clock, so the outer `nonClickableMsg` (`"Element is not clickable after Xs"`) is unreachable in practice.

Suggested fix — remove the inner `waitUntil` entirely and let the outer loop retry `getAllDisplayed` on each iteration:

```ts
await browser.waitUntil(
  async () => {
    let elems: Array<Element> = [];
    try {
      elems = await ui5.element.getAllDisplayed(selector, timeout);
    } catch {
      return false;
    }
    const isClickableFlags = await Promise.all(elems.map((e) => e.isClickable()));
    elem = elems.filter((_, i) => isClickableFlags[i])[index];
    return !!elem;
  },
  { timeout, interval: GLOBAL_DEFAULT_WAIT_INTERVAL, timeoutMsg: nonClickableMsg }
);
```

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -729,71 +729,36 @@ export class UserInteraction {
...
+            interval: GLOBAL_DEFAULT_WAIT_INTERVAL,
+            timeoutMsg: timeoutMsg
+          }
+        );
+        // Non-clickable elements are not counts while indexing
+        const isClickableFlags = await Promise.all(elems.map(async (e) => e.isClickable()));
+        elem = elems.filter((_, i) => isClickableFlags[i])[index];
```
DIFSRIP: **[Required]** Silent API-breaking change: `index` previously addressed the Nth element among all *displayed* elements (the old code called `getDisplayed(selector, index, timeout)` which applied the index before returning). Now it addresses the Nth among only the *clickable* ones — non-clickable elements are filtered out before `[index]` is applied.

Any caller passing `index > 0` where some preceding elements are temporarily disabled (e.g. "click the 2nd button, the 1st is greyed out") will now silently receive a different element than intended, with no error thrown.

Please document this semantic change explicitly in:
- The PR description
- The JSDoc for the three public callers: `click` (L40), `doubleClick` (L76), and `rightClick` (L97)

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -729,71 +729,36 @@ export class UserInteraction {
...
-        return (parentElement ? await resolveZIndex(parentElement) : 0);
-      }
-      return parseInt(zIndex);
-    }
-
-    const poll = async () => {
-      elem = await ui5.element.getDisplayed(selector, index, timeout);
-      return !!(elem && (await elem.isClickable()));
-    };
+    const nonClickableMsg = `Element is not clickable after ${timeout / 1000}s`;
```
DIFSRIP, line 731: **[Recommended]** The two variable names are semantically inverted relative to their contents:
- `nonClickableMsg` (this line) holds `"Element is not clickable after Xs"` — used as the **outer** `waitUntil`'s `timeoutMsg` (L758)
- `timeoutMsg` (next line) holds `"Not found elements with selector:..."` — used as the **inner** not-found message (L746)

A reader scanning the error-handling block expects `timeoutMsg` to be the final timeout error. Suggested rename:

```ts
const nonClickableMsg = `Element is not clickable after ${timeout / 1000}s`;
const notFoundMsg     = `No elements found with selector: ${JSON.stringify(selector)}`;
```

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -729,71 +729,36 @@ export class UserInteraction {
...
+    const timeoutMsg = `Not found elements with selector: ${JSON.stringify(selector)}`;
```
DIFSRIP: **[Optional]** Grammar: `"Not found elements with selector:..."` reads awkwardly. The old codebase used the pattern `"No <X> found with selector:..."`. Consistent suggestion:

```ts
`No elements found with selector: ${JSON.stringify(selector)}`
```

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -729,71 +729,36 @@ export class UserInteraction {
...
+            elems = await ui5.element.getAllDisplayed(selector, timeout);
+            return elems.length !== 0;
+          },
+          {
+            timeout: timeout,
+            interval: GLOBAL_DEFAULT_WAIT_INTERVAL,
+            timeoutMsg: timeoutMsg
+          }
+        );
+        // Non-clickable elements are not counts while indexing
```
DIFSRIP: **[Optional]** Typo: `"are not counts"` → `"are not counted"`.

test/reuse/ui5/userInteraction/doubleClick.spec.js
```diff
@@ -65,6 +65,6 @@ describe("userInteraction - doubleClick on invisible element", function () {
     const timeout = 30000;
     await ui5.userInteraction.doubleClick(selector, index, timeout);
     await browser.pause(1000);
-    await expect(ui5.userInteraction.doubleClick(selector, index, timeout)).rejects.toThrow(/No clickable elements found/);
+    await expect(ui5.userInteraction.doubleClick(selector, index, timeout)).rejects.toThrow(/Element is not clickable after 30s/);
```
DIFSRIP: **[Recommended]** The regex `/Element is not clickable after 30s/` hardcodes the timeout duration. If the `timeout` constant on L65 is ever changed, this assertion silently stops matching the actual error message — the test passes vacuously rather than failing with a clear mismatch.

The parallel test in `click.spec.js:72` already handles this correctly:
```js
.rejects.toThrow(/Element is not clickable after/)
```
Apply the same pattern here (drop the `30s`).

test/reuse/ui5/userInteraction/click.spec.js
```diff
@@ -98,23 +102,37 @@ describe("userInteraction - click unblocked button and ignore blocked one", func
       text: "Create"
     }
   };
+  const assertSelector = {
+    elementProperties: {
+      metadata: "sap.m.Title",
+      id: "__field2-sfEdit-input-valueHelpDialog-title"
```
DIFSRIP: **[Optional]** `__field2-sfEdit-input-valueHelpDialog-title` starts with `__`, which is UI5's auto-generated ID prefix — these IDs are not guaranteed to be stable across app updates or rendering cycles. If this ID changes, both the `Preparation` step (L122) and the `Verification` step (L136) will silently time out rather than fail with a clear "element not found" error.

If the value help dialog title has a stable `text` property or a custom stable ID, prefer matching on that instead.

DIFSRIP (dismissed):
This refactor significantly simplifies `_getClickableElement` and removes the fragile z-index overlay detection in favour of the W3C-spec `isClickable()` check — the right long-term direction. The Dependabot config, `CHROMEDRIVER_PATH` fix, and `environment: test` scoping are all solid additions.

Two issues need to be resolved before merging: a nested `waitUntil` timing bug that makes the outer timeout unreachable, and a silent breaking change to `index` semantics. The remaining items are polish.

**Dead code (cleanup):** `OVERLAY_CHECK_TIMEOUT = 5000` at `src/reuse/modules/ui5/userInteraction.ts:25` is no longer referenced after the refactor — it was used exclusively by the removed two-phase polling logic. Remove it.

test/reuse/ui5/userInteraction/doubleClick.spec.js
imaroz-lx: Done

test/reuse/ui5/userInteraction/click.spec.js
imaroz-lx: Yes, there is a stable text. Replaced.

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Not relevant anymore

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Replaced with rethrow default uiControls error message.

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Refactored

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Inner wait loop removed.

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Thanks for that catch, this is actually quite tricky and make me spends some days to find better solution.
- From one hand when we work with multiple elements and part of them are disabled we need to specify desired index.
- From another when we work in dialogs we don't care about elements outside and provided index works only in scope of active dialog.

In summary: we can't let users index elements outside active dialog. But if we filter out all blocked elements and just update docs this way we create 2 new problems:
- logic error when user intentionally wants to fails while clicking blocked element.
- it may (read as "it 90% will") leads to breaking multiple existing tests

But. Surprisingly all this time we have hidden ability to search descendant elements in already found one (`Element.uiControl`). For some reason this was not ins use and even not in types.

So in final solution here we trying to find an active dialog and then use it as a root element for our selectors. If no dialogs -> proceed with default logic.

test/reuse/ui5/userInteraction/click.spec.js
```diff
@@ -91,30 +95,64 @@ describe("userInteraction - click with selector having wildcard character(*) for
...
+          tooltipLabel: "EPM: Total Gross Amount"
+        }
+      }
+    };
+    await ui5.userInteraction.click(valueHelpIconSelector);
+    await ui5.assertion.expectToBeVisible(assertSelector);
   });
 
   it("Execution", async function () {
+    const selector = {
```
uladhsi: [Optional]: rename const to `createButtonSelector`

test/reuse/ui5/userInteraction/click.spec.js
```diff
@@ -91,30 +95,64 @@ describe("userInteraction - click with selector having wildcard character(*) for
...
+    await browser.navigateTo(`${BASE_URL}/test-resources/sap/suite/ui/generic/template/demokit/demokit.html?responderOn=true&demoApp=sttasalesordertt#/?sap-iapp-state=3&sap-iapp-state--history=1`);
+  });
+
+  it("Execution && Verification", async function () {
+    await expect(ui5.userInteraction.click(selector, 0, NEGATIVE_TEST_TIMEOUT)).rejects.toThrow(/Element is not clickable after/);
+  });
+});
+
 describe("userInteraction - click unblocked button and ignore blocked one", function () {
   const selector = {
```
uladhsi: [Optional]: rename this const to `createButtonSelector`

test/reuse/ui5/userInteraction/click.spec.js
```diff
@@ -91,30 +95,64 @@ describe("userInteraction - click with selector having wildcard character(*) for
...
     elementProperties: {
       metadata: "sap.m.Button",
       text: "Create"
     }
   };
+  const assertSelector = {
```
uladhsi: [Optional]: rename this const to `currencyCodeSelector`

uladhsi (changes requested):
Just a few comments

test/reuse/ui5/userInteraction/click.spec.js
imaroz-lx: Done

test/reuse/ui5/userInteraction/click.spec.js
imaroz-lx: Done

test/reuse/ui5/userInteraction/click.spec.js
imaroz-lx: Done

uladhsi (dismissed):

DIFSRIP (dismissed):

imaroz-lx:
New changes! Due to previous implementation does not consider popups inside of the dialog now click available for the latest general popup scope.

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -728,73 +727,58 @@ export class UserInteraction {
...
-      // intentionally continue with additional checks instead of throwing.
-      if (!isQuickPhase || e?.message !== QUICK_TIMEOUT_MSG) {
-        throw e;
-      }
+      await browser.waitUntil(
+        async () => {
+          try {
+            // Warn. It is important to use elements only from active dialogs and inner popups when they presented
+            const activePopups = await this._getActiveDialogPopups();
+            if (activePopups) {
```
DIFSRIP, line 742: Optional: Maybe use an early return here, to reduce the nesting

src/reuse/modules/ui5/userInteraction.ts
```diff
@@ -728,73 +727,58 @@ export class UserInteraction {
...
+        .filter((e) => e.checkVisibility())
+        .flatMap((e) => sap.ui.getCore().byId(e.id) ?? [])
+        .filter((e) => e?.getMetadata()?.isInstanceOf("sap.ui.core.PopupInterface"));
+      const lastDialog = popups.findLastIndex((p) => p.isA("sap.m.Dialog"));
+      return lastDialog < 0 ? undefined : popups.slice(lastDialog).map((p) => p.getId());
     });
-    return elem!;
+
+    if (!Array.isArray(activePopupsIds) || activePopupsIds.length == 0) return undefined;
+    vl.log(`Found an opened popups with ids: ${JSON.stringify(activePopupsIds)}`);
```
DIFSRIP: Small grammer fix:  
```suggestion
    vl.log(`Found open popups with the following IDs: ${JSON.stringify(activePopupsIds)}`);
```

DIFSRIP (changes requested):
Looks good, just a small grammer fix

src/reuse/modules/ui5/userInteraction.ts
imaroz-lx: Sorry, don't see where can be applied an early return.

When there is an active dialogs on the page one flow executed. Otherwise used standard getDisplayed approach.

DIFSRIP: My bad, the early return pattern does not apply here, as both paths assign elems.

DIFSRIP (dismissed):
LGTM

cvakiitho (approved):

PR #570, opened 2026-06-02
chore: add Dependabot config with supply-chain hardening

cvakiitho:
Adds .github/dependabot.yml covering npm and github-actions ecosystems.

Hardening against npm supply-chain attacks (e.g. Shai-Hulud, Sept 2025):
- 7-day cooldown on minor/patch, 14-day on major version updates so the registry/community have time to detect and unpublish malicious releases before Dependabot proposes them. Cooldown is bypassed for security advisories so CVE/GHSA fixes still flow through immediately.
- Updates grouped by dependency-type and semver impact so breaking changes are never silently bundled with patches.
- Security updates get a dedicated group so advisories aren't buried in large grouped PRs.
- github-actions ecosystem keeps the existing SHA pins fresh, defending against tag-rewriting attacks (cf. tj-actions/changed-files, 2025).
- versioning-strategy: increase keeps semver ranges in package.json intact and only bumps on real new versions.

WebdriverIO is pinned to the v8 line via ignore rules on @wdio/* and wdio-* (major updates suppressed; minor/patch and security advisories still flow). Remove the ignore block when ready to migrate to v9.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,132 @@
+# Dependabot configuration
+# Reference: https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+#
+# Hardening rationale (supply-chain defence — e.g. "Shai-Hulud" npm worm, Sept 2025):
+#   - `cooldown` delays PRs for newly published versions so the npm registry and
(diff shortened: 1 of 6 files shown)


DIFSRIP (approved):
Looks good to me. Thanks for the ifx and improvements!

PR #571, opened 2026-06-02
fix: do not reuse odata service instance if username or password changes in flp.userSettings.getDateFormatFromUserSettings and other user settings functions

heraldantony:
If user calls `flp.userSettings.getDateFormatFromUserSettings` with two different sets of usernames and passwords, the second call to this function returns date format for the first user.
This is because we initialize odata service with first user credentials, and retrieve user date settings with the same service instance in all subsequent calls - https://github.com/SAP/wdio-qmate-service/blob/e331e6dc9e75caa4f267126a7c73b8566ed264dd/src/reuse/modules/flp/userSettings.ts#L30

Fixed the code, so that if username or password is changed between calls, the odata service is re-initialized.

### DIFF
--- a/src/reuse/modules/flp/userSettings.ts
+++ b/src/reuse/modules/flp/userSettings.ts
@@ -10,6 +10,8 @@ import { VerboseLoggerFactory } from "../../helper/verboseLogger";
 export class UserSettings {
   private vlf = new VerboseLoggerFactory("util", "userSettings");
   private _srvInstance = null;
+  private _user: string = "";
+  private _password: string = "";
 
   /**
    * @private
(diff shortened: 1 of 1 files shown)


src/reuse/modules/flp/userSettings.ts
```diff
@@ -27,12 +29,15 @@ export class UserSettings {
       throw new Error(`${emptyField} is empty, undefined or null. Please provide valid credentials for S4 User Setting Service initialization.`);
     }
 
-    if (!this._srvInstance) {
+    // if username or password changes re-initialize odata service instance
```
uladhsi:
```suggestion
    // If service instance is not initialized, OR username or password changes --> re-initialize odata service instance
```

uladhsi (changes requested):
one small suggestion, everything else looks good

src/reuse/modules/flp/userSettings.ts
heraldantony: thanks Vlad, updated comment

uladhsi (approved):
Looks good to me

PR #593, opened 2026-06-25
Chore/update pdf parse

cvakiitho:


cvakiitho:
I don't really like the canvas dependency inclusion, will try to work on some other solution

uladhsi (approved):

test/reuse/util/file/pdfParser.spec.js
```diff
@@ -28,8 +28,7 @@ describe("PDF Parser", function () {
   describe("expectPdfContainsText - from URL", function () {
     it("Execution and Verification", async function () {
       const pdfPath = "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf";
-      // await util.file.expectPdfContainsText(pdfPath, "Dummy PDF file");
-      await util.file.expectPdfContainsText(pdfPath, "Dumm y   PDF  fi le");
+      await util.file.expectPdfContainsText(pdfPath, "Dummy PDF file");
```
cvakiitho, line 31: just a note, this change, fixes formatting issues, so it might be causing some false positive failures in already existing tests, so probably a good idea to keep that in mind in release notes.

PR #594, opened 2026-06-25
chore: remove unused mobile reuse library and appium deps (0 usage)

cvakiitho:


uladhsi (approved):
Looks good to me, but I would double check with @DIFSRIP that we really don't need that namespace anywhere

DIFSRIP (approved):
Looks good to me. 

We can re-introduce certain functionality if required, but I guess no one is using/demanding it at the moment.

PR #606, opened 2026-07-16
feat: support btp logout text

uladhsi:


DIFSRIP (dismissed):
Implementation looks good so far but I suggest to update the description to reflect the change telling that it supports different systems. Additionally, as we already touch the function, we could introduce some timeout parameter to the arguments? And use the default as fallback if not passed. What do you think?

src/reuse/modules/ui5/session.ts
```diff
@@ -271,14 +271,51 @@ export class Session {
...
+          }
+        },
+        {
+          timeout: timeout,
+          timeoutMsg: `Logout text not visible in ${+timeout / 1000}s`,
+          interval: GLOBAL_DEFAULT_WAIT_INTERVAL
+        }
+      );
+    } catch (error) {
+      this.ErrorHandler.logException(error);
```
uaroragit: try/catch swallows errors — unlike the original implementation which let errors propagate naturally. Either remove the try/catch entirely and let browser.waitUntil throw on its own (which it does by default), or add throw error after the log. 

I think the simplest fix is to remove the outer try/catch completely.

src/reuse/modules/ui5/session.ts
```diff
@@ -271,14 +271,51 @@ export class Session {
   /**
    * @function expectLogoutText
    * @memberOf ui5.session
-   * @description Expects the logout text after logout to be "You have been logged off.
+   * @description Expects the logout text after logout. Waits for "You have been logged off." in S4 systems or "Goodbye" in BTP systems.
    * This is essential for chaining scripts, so that no static browser sleep in the spec itself is required anymore.
+   * @param {Number} [timeout=30000] - The timeout to wait (ms).
```
uaroragit:
```suggestion
* @param {Number} [timeout=30000] - The timeout in ms (default: 30000ms or QMATE_CUSTOM_TIMEOUT env var)
```

uaroragit (changes requested):

src/reuse/modules/ui5/session.ts
uladhsi: done: https://github.com/SAP/wdio-qmate-service/pull/606/commits/2c84923b6e3aed03dea38a88c1a35c8fdc2fac3c

src/reuse/modules/ui5/session.ts
uladhsi: done: https://github.com/SAP/wdio-qmate-service/pull/606/commits/4d4daa519c18ec5b91ea1ada4b73541d83e180e5

uaroragit (approved):

PR #640, opened 2026-09-09
fix: update selectors for failing specs

uladhsi:
Update specs which fail due to the recent SAPUI5 update to 1.152.0


uaroragit (approved):
Looks good!

PR #641, opened 2026-09-11
fix: ui5.session.logout

imaroz-lx:


DIFSRIP (dismissed):
LGTM

src/reuse/modules/ui5/navigationBar.ts
```diff
@@ -85,65 +85,45 @@ export class NavigationBar {
...
-          metadata: "sap.f.gen.ui5.webcomponents.dist.Avatar",
-          id: "userActionsMenuHeaderButton"
+          metadata: "sap.*.Avatar",
+          id: "*HeaderButton"
         }
       };
-      await ui5.userInteraction.click(selector, 0, 500);
+      const id = await ui5.element.getId(selector, 0, iterationTimeout);
+      await util.browser.executeScript((id: string) => {
+        sap.ui.getCore().byId(id).firePress();
```
DIFSRIP, line 109: Please add some inline comments next to the command to describe the need for this hack

src/reuse/modules/ui5/navigationBar.ts
```diff
@@ -85,65 +85,45 @@ export class NavigationBar {
...
       await browser.waitUntil(
         async () => {
-          try {
-            await Promise.any([clickLegacyUserAvatar(), clickWebComponentUserProfile(), clickShellBarUserAvatar(), clickShellBarUserAvatar2()]);
-            return true;
-          } catch (error) {
-            // Ignore error and continue to next promise
-            return false;
-          }
+          const results = await Promise.allSettled([clickWebComponentUserProfile(), clickShellBarUserAvatar()]);
```
DIFSRIP, line 116: Just for curiosity, what was the need to switch from Promise.any to Promise.allSettled?

DIFSRIP (changes requested):
Just a minor thing.

imaroz-lx:
https://github.tools.sap/sProcurement/qmate/issues/1299

src/reuse/modules/ui5/navigationBar.ts
imaroz-lx: Outdated

src/reuse/modules/ui5/navigationBar.ts
imaroz-lx: Outdated. 
P.S. Originally, to handle errors correctly

src/reuse/modules/ui5/navigationBar.ts
```diff
@@ -85,66 +85,32 @@ export class NavigationBar {
...
-          timeout: timeout,
-          timeoutMsg: `Could not click User Icon in ${+timeout / 1000}s`,
-          interval: GLOBAL_DEFAULT_WAIT_INTERVAL
-        }
-      );
-    } catch (error) {
-      this.ErrorHandler.logException(error);
-    }
+    await Promise.any([clickWebComponentUserProfile(), clickShellBarUserAvatar()]).catch((e) => {
+      throw new Error(`Could not click User Icon in ${+timeout / 1000}s: ${(e as AggregateError).errors}`);
```
DIFSRIP: I guess we need to join the messages specifically since the might be stringified to [object Error]
```suggestion
      const reasons = (e as AggregateError).errors.map((err: Error) => err.message).join(", ");
      throw new Error(`Could not click User Icon in ${+timeout / 1000}s: ${reasons}`);
```

src/reuse/modules/ui5/session.ts
```diff
@@ -231,9 +231,28 @@ export class Session {
...
+          // Wait a little for a popup to stabilize (prevents instant closing)
+          await util.browser.sleep(200);
+          await this._clickSignOut(actionTimeout);
+          vl.log("Clicking OK to confirm logout...");
+          await ui5.confirmationDialog.clickOk(actionTimeout);
+          return true;
+        } catch (e) {
+          return ((lastError = e), false);
+        }
+      });
```
DIFSRIP:
```suggestion
     }, { timeout: GLOBAL_DEFAULT_WAIT_TIMEOUT, timeoutMsg: `Logout flow did not complete in ${GLOBAL_DEFAULT_WAIT_TIMEOUT / 1000}s` });
```

src/reuse/modules/ui5/session.ts
```diff
@@ -377,34 +401,21 @@ export class Session {
...
-        }
-      },
-      {
-        timeout: timeout,
-        timeoutMsg: `Could not click Sign out button in ${+timeout / 1000}s`,
-        interval: GLOBAL_DEFAULT_WAIT_INTERVAL
-      }
-    );
+    await Promise.any([scrollAndClickLogoutOld(), scrollAndClickLogoutNew()]).catch((e) => {
+      throw new Error(`Could not click Sign out button in ${+timeout / 1000}s: ${(e as AggregateError).errors}`);
```
DIFSRIP: Without this, ${errors} stringifies the array as Error: msg1,Error: msg2 . The Error: prefix is noise and there's no space after the comma. Extracting .message and joining keeps it readable.
```suggestion
      const reasons = (e as AggregateError).errors.map((err: Error) => err.message).join(", ");
      throw new Error(`Could not click Sign out button in ${+timeout / 1000}s: ${reasons}`);
```

src/reuse/modules/ui5/session.ts
```diff
@@ -279,40 +298,45 @@ export class Session {
...
-        try {
-          await Promise.any([expectS4LogoutText(), expectBtpLogoutText()]);
+    let lastError: unknown;
+    try {
+      await browser.waitUntil(
+        async () => {
+          try {
+            await Promise.any([isS4LogoutTextVisible(), isBtpLogoutTextVisible()]);
+          } catch (e) {
+            return ((lastError = (e as AggregateError).errors), false);
```
DIFSRIP: I believe `isVisible `always resolves with a boolean, never rejects. So `Promise.any` fulfills with false when neither is visible, the catch is never hit, and the callback returns true. Hence, the loop exits immediately thinking it's done. 

Maybe we can switch to Promise.all + explicit check?:
```suggestion
    try {
      await browser.waitUntil(
        async () => {
          const [s4Visible, btpVisible] = await Promise.all([
            isS4LogoutTextVisible().catch(() => false),
            isBtpLogoutTextVisible().catch(() => false)
          ]);
          if (!s4Visible && !btpVisible) {
            lastError = new Error("Logout text not visible in S4 or BTP");
            return false;
```

DIFSRIP (changes requested):
Thanks for fixing it! 
Just some minor comments and one clarification needed.

src/reuse/modules/ui5/session.ts
imaroz-lx: Good catch. However suggested approach is complex and mostly repeats one before fixes.
So I reverted back to old one but with only change is resolved timeout issues.

DIFSRIP: So it works now as expected?

DIFSRIP (dismissed):

DIFSRIP (approved):

src/reuse/modules/ui5/session.ts
imaroz-lx: Discussed, so reverted changes are leaved, only timeout fixes applied.

PR #642, opened 2026-09-16
fix: nonUi5 click

imaroz-lx:
https://github.tools.sap/sProcurement/qmate/issues/1301

PR #643, opened 2026-09-17
review: improve nonUi5.userInteraction.click robustness and error clarity

DIFSRIP:
## Summary

Review suggestions for #642. All changes are confined to the `click` function in `nonUi5/userInteraction.ts`.

- **Cache resolved element** — `resolveCssSelectorOrElement` is now only called once. Subsequent `waitUntil` iterations skip straight to `isDisplayed`/`isEnabled` re-checks, avoiding redundant resolution and inaccurate per-iteration timeout errors in logs.
- **Rename `iterationTimeout` → `resolveTimeout`** — the capped timeout only applies to the resolution call, not to display/enabled checks.
- **Prevent fall-through to click block on timeout** — added `return` after the outer `catch` so a `waitUntil` timeout no longer silently falls through to `element!.click()` on an unresolved element.
- **Distinguish failure scenarios in error messages** — if `element` is still `undefined` at timeout, logs a message referencing the full `timeout` value; if the element was found but never became ready, logs the specific `lastError`.
- **Include selector in all error messages** — `isDisplayed`, `isEnabled`, and click failure errors now all reference the selector or element for easier diagnosis.
- **Separate highlight and click try/catch** — a highlight failure no longer suppresses the click. Highlight errors are logged and the click proceeds.
- **Type safety** — `lastError` typed as `| undefined`, comma expression replaced with explicit statements, `lastError ?? (e as Error)` fallback instead of `!` assertion.
- **JSDoc** — updated description and `timeout` param to reflect the actual waiting behaviour.

## Test plan

- [ ] Verify click succeeds when element is initially not present, then appears within timeout
- [ ] Verify element is not re-resolved after initial resolution (e.g. log output shows resolve called once)
- [ ] Verify timeout error message references the full timeout and selector when element never resolves
- [ ] Verify click still proceeds when highlight throws
- [ ] Verify click failure error includes selector and underlying reason


src/reuse/modules/nonUi5/userInteraction.ts
```diff
@@ -23,44 +23,63 @@ export class UserInteraction {
...
       await browser.waitUntil(
         async () => {
           try {
-            element = await resolveCssSelectorOrElement(elementOrSelector, iterationTimeout);
-            if (!(await element.isDisplayed())) throw new Error("Element is found, but not displayed");
-            if (!(await element.isEnabled())) throw new Error("Element is found, displayed, but not enabled");
+            // Only resolve once; re-checking isDisplayed/isEnabled is sufficient on subsequent iterations.
+            if (!element) {
+              element = await resolveCssSelectorOrElement(elementOrSelector, resolveTimeout);
+            }
```
imaroz-lx: Two concerns here:
- This check would prevents from getting most fresh element from page. Example:
Some list/container loads rows sequentially and gets busy while do that (row are disabled/invisible until full load) -> first row loaded? -> locked with this check -> more new rows loaded? -> ignored, wait for initial row is disabled -> click that row even if that no the first element -> user confused )
- This check eventually would suppress retrieve payload (leaving almost instant isDisplayed, isEnabled checks uncontrollable at hundreds per second execution rate), polling interval should be provided.

src/reuse/modules/nonUi5/userInteraction.ts
```diff
@@ -23,44 +23,63 @@ export class UserInteraction {
...
         { timeout }
       );
     } catch (e) {
-      this.ErrorHandler.logException(lastError!);
+      if (!element) {
+        this.ErrorHandler.logException(new Error(`Element could not be resolved within ${timeout}ms. Last error: ${lastError?.message}`));
+      } else {
+        this.ErrorHandler.logException(lastError ?? (e as Error));
+      }
+      return;
```
imaroz-lx: Unreachable code

src/reuse/modules/nonUi5/userInteraction.ts
```diff
@@ -23,44 +23,63 @@ export class UserInteraction {
...
+            lastError = e as QmateError | Error;
+            return false;
           }
         },
         { timeout }
       );
     } catch (e) {
-      this.ErrorHandler.logException(lastError!);
+      if (!element) {
+        this.ErrorHandler.logException(new Error(`Element could not be resolved within ${timeout}ms. Last error: ${lastError?.message}`));
```
imaroz-lx: Is there a specific reason to override internal `resolveCssSelectorOrElement` error?

P.S. Last error here would be always error from `resolveCssSelectorOrElement`

src/reuse/modules/nonUi5/userInteraction.ts
```diff
@@ -23,44 +23,63 @@ export class UserInteraction {
...
     }
 
     try {
       vl.log("Clicking the element");
-      if (highlightConfig.enable) await nonUi5.element.highlight(element!, highlightConfig.duration, highlightConfig.color);
       await element!.click();
     } catch (error) {
-      this.ErrorHandler.logException(error);
+      const selector = typeof elementOrSelector === "string" ? elementOrSelector : "provided element";
+      this.ErrorHandler.logException(new Error(`Failed to click element with selector '${selector}'. Reason: ${(error as Error).message}`));
```
imaroz-lx:
```suggestion
      const isSelector =  typeof elementOrSelector === 'string';
      this.ErrorHandler.logException(new Error(`Failed to click ${isSelector ? `element with selector ${selector}` : "provided element"} Reason: ${(error as Error).message}`));
```

imaroz-lx (changes requested):

src/reuse/modules/nonUi5/userInteraction.ts
DIFSRIP: I agree

DIFSRIP: Good points on both counts. Removed the caching guard — `resolveCssSelectorOrElement` now re-runs every iteration, so the element reference is always fresh and dynamic DOM changes (list rows, re-renders) are handled correctly. Also added `interval: GLOBAL_DEFAULT_WAIT_INTERVAL` to `waitUntil` to make the polling rate explicit rather than relying on the WebdriverIO default of 500ms. Note that since `resolveCssSelectorOrElement` itself takes time when the element is not yet present, the effective cycle duration will naturally be higher than the interval alone.

src/reuse/modules/nonUi5/userInteraction.ts
DIFSRIP: Agreed — if `logException` always throws (as it does in this framework), the `return` is unreachable. Removed it along with the `!element` branch in the catch, simplifying the outer catch to a single `logException(lastError ?? (e as Error))`.

src/reuse/modules/nonUi5/userInteraction.ts
DIFSRIP: The reason was to avoid a misleading error message: `resolveCssSelectorOrElement` is called with `resolveTimeout` (capped at 3s), so its own error would reference 3s — but the user actually waited the full `timeout` (e.g. 30s) across all iterations. Wrapping it in a new message with the total `timeout` value makes the failure clearer in test logs. That said, after removing the element caching (see comment above), this distinction is less important since the outer catch is now a simple passthrough of `lastError`.

src/reuse/modules/nonUi5/userInteraction.ts
DIFSRIP: Good suggestion, applied. Used `isSelector` as the variable name and kept the quotes around the selector value for readability.

imaroz-lx (approved):

